PR #48, opened 2022-06-22
Add a hack UI for Grafana Phlare

simonswine:
This is a follow-up on #47 to demonstrate how to use the openapiv2 specs.

It adds a very basic UI that can be extended and matches approach of prometheus.


simonswine:
This has been superseeded by a real UI :joy:

PR #388, opened 2022-11-02
Fix README

simonswine:
- Hide play.grafana.com, while it is not working there
- Fix link to Grafana Phlare deploy


hairyhenderson (approved):
:shipit:

PR #393, opened 2022-11-03
[Helm] Load environment variables from secrets

nlamirault:
Signed-off-by: Nicolas Lamirault <nicolas.lamirault@gmail.com>

### DIFF
--- a/operations/phlare/helm/phlare/templates/deployments-statefulsets.yaml
+++ b/operations/phlare/helm/phlare/templates/deployments-statefulsets.yaml
@@ -65,6 +65,10 @@ spec:
               value: {{ $value | quote }}
            {{- end }}
           {{- end }}
+          {{- with .Values.extraEnvFrom }}
+          envFrom:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           ports:
             - name: {{ $values.service.port_name }}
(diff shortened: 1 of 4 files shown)


operations/phlare/helm/phlare/templates/deployments-statefulsets.yaml
```diff
@@ -65,6 +65,10 @@ spec:
               value: {{ $value | quote }}
            {{- end }}
           {{- end }}
+          {{- with .Values.extraEnvFrom }}
```
simonswine, line 68: We allow a couple of overrides to be happening and hence you need to access the values like this:

```suggestion
          {{- with .$values.extraEnvFrom }}
```

simonswine (commented):
Thanks for your contribution! 

I think the way it is, it won't work as you won't access the correct values, see my specific comment.

Other than that: 
* Could you please rebase with the latest main (to fix the broken tests)?
* If possible run `make helm/check` and commit the changed file (we really should have a CI check for that #396 )

simonswine (approved):
Perfect! That's ready to go. Thanks for the contribution

nlamirault:
@simonswine any news to release a new chart version with this feature ?

nlamirault:
Thanks @simonswine

Issue #410, opened 2022-11-06
agent ignores 'path_prefix'


kevinschoonover:
#### Describe the bug
Agent does not prepend the 'profiling_config' argument when scraping targets

```yml
scrape_configs:
- job_name: vault
  authorization:
    credentials_file: "/secrets/vault_token"
  scheme: https
  profiling_config:
    path_prefix: '/v1/sys/pprof'
  tls_config:
    insecure_skip_verify: true
  static_configs:
    - targets:
        - 100.114.243.60:8200
        - 100.93.8.11:8200
        - 100.111.89.83:8200
        

      labels:
        env: prod
```

```
evel=error caller=target.go:183 ts=2022-11-06T08:58:19.92511409Z msg="fetch profile failed" target="{__address__=\"100.111.89.83:8200\", __name__=\"process_cpu\", __profile_path__=\"/debug/pprof/profile\", __scheme__=\"https\", env=\"prod\", instance=\"100.111.89.83:8200\", job=\"vault\"}" err="server returned HTTP status (404) {\"errors\":[]}"
level=error caller=target.go:183 ts=2022-11-06T08:58:22.352563943Z msg="fetch profile failed" target="{__address__=\"100.93.8.11:8200\", __name__=\"block\", __profile_path__=\"/debug/pprof/block\", __scheme__=\"https\", env=\"prod\", instance=\"100.93.8.11:8200\", job=\"vault\"}" err="server returned HTTP status (404) {\"errors\":[]}"
level=error caller=target.go:183 ts=2022-11-06T08:58:22.770612925Z msg="fetch profile failed" target="{__address__=\"100.114.243.60:8200\", __name__=\"mutex\", __profile_path__=\"/debug/pprof/mutex\", __scheme__=\"https\", env=\"prod\", instance=\"100.114.243.60:8200\", job=\"vault\"}" err="server returned HTTP status (404) {\"errors\":[]}"
level=error caller=target.go:183 ts=2022-11-06T08:58:23.198404024Z msg="fetch profile failed" target="{__address__=\"100.93.8.11:8200\", __name__=\"mutex\", __profile_path__=\"/debug/pprof/mutex\", __scheme__=\"https\", env=\"prod\", instance=\"100.93.8.11:8200\", job=\"vault\"}" err="server returned HTTP status (404) {\"errors\":[]}"
[...]

PR #412, opened 2022-11-07
Fixes scrape_config path_prefix.

cyriltovena:
The path prefix was not appended correctly during valition. This PR also upgrade some dependencies.

Fixes https://github.com/grafana/phlare/issues/410


simonswine (approved):
LGTM

I think the default values are still not idea, to configure it successfully you need to add this to the config (delta as otherwise there is not seconds parameter, and the default path is wrong `debug/pprof/fgprof`, but it should be `debug/fgprof`) :

```diff
diff --git a/cmd/phlare/phlare.yaml b/cmd/phlare/phlare.yaml
index 52a1c408..6293a38a 100644
--- a/cmd/phlare/phlare.yaml
+++ b/cmd/phlare/phlare.yaml
@@ -6,3 +6,9 @@ scrape_configs:
     scrape_interval: "15s"
     static_configs:
       - targets: ["127.0.0.1:4100"]
+    profiling_config:
+      pprof_config:
+        fgprof:
+          enabled: true
+          delta: true
+          path: /debug/fgprof
```

simonswine:
Also not entirely sure if the units are correct (they are the same though in pprof):

scrn-2022-11-23-15-10-01

PR #413, opened 2022-11-07
Adds fgprof to Phlare

cyriltovena:
This enabled fgprof which allowed me to verify that we can correctly parse those profiles.


simonswine (approved):
LGTM

PR #414, opened 2022-11-08
Upgrades to jpprof 0.1.4

cyriltovena:
This also uses seccomp instead of `CAP_SYS_ADMIN` to advertise better security management.


cristiangreco (approved):
Nice!

PR #424, opened 2022-11-21
Ensure opening block is atomic

cyriltovena:
Previously if one of the tables open would fail, the block would be considered open and leading to panic.


simonswine (approved):
Good catch

LGTM

Issue #421, opened 2022-11-16
Panic internal: unknown: error parquet file 'mappings.parquet' contains no rows


mattiaforc:
#### Describe the bug
I managed to run phlare once, actually scraping a spring boot application running on my laptop; after restarting the containers a panic shows in phlame logs and interface won't show a thing anymore.

#### To Reproduce

I started off the docker-compose example on branch master, modified both `docker-compose.yaml` and `phlare.yaml` as follows:
`phlare.yaml`:
```yaml
scrape_configs:
  - job_name: "java"
    scrape_interval: "15s"
    static_configs:
      - targets: ["localhost:8012"]
    profiling_config:
      pprof_config:
        block: { enabled: false }
        goroutine: { enabled: false }
        memory: { enabled: false }
        mutex: { enabled: false }
```
`docker-compose.yml`:
```yaml
services:
  phlare:
    image: grafana/phlare:latest
    command: -config.file=/etc/phlare/config.yaml
    volumes:
      - ./phlare.yaml:/etc/phlare/config.yaml
      - data:/data
    network_mode: host

  grafana:
    image: grafana/grafana:main
    environment:
      - GF_FEATURE_TOGGLES_ENABLE=flameGraph
      - GF_AUTH_ANONYMOUS_ENABLED=true
      - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
      - GF_DIAGNOSTICS_PROFILING_ENABLED=true
      - GF_DIAGNOSTICS_PROFILING_ADDR=0.0.0.0
      - GF_DIAGNOSTICS_PROFILING_PORT=6060
    volumes:
      - ./datasource.yaml:/etc/grafana/provisioning/datasources/datasources.yml
    network_mode: host

volumes:
  data:

    # yaml-language-server: $schema=https://raw.githubusercontent.com/compose-spec/compose-spec/master/schema/compose-spec.json
```
Then I followed the [quickstart for JVM](https://grafana.com/docs/phlare/latest/operators-guide/configure-agent/language-support/jvm/) and added the profiling endpoint to my Spring Boot application.

I ran `docker-compose up -d` the first time and everything went fine, I could see my "java" profiling correctly. I shut both grafana and phlame down with `docker-compose down` and added the following to `phlame.yaml` (as seen in the example):
```yaml
  - job_name: "phlare"
    scrape_interval: "15s"
    static_configs:
      - targets: ["localhost:4100"]
  - job_name: "grafana"
    scrape_interval: "15s"
[...]

PR #426, opened 2022-11-21
Ensure we have at least one mapping in pprof

cyriltovena:
Fixes https://github.com/grafana/phlare/issues/421

This is following along the line of https://github.com/google/pprof/blob/e6195bd50e260a275f32e5bd46f079a3abae0f9e/profile/legacy_profile.go#L190


simonswine (approved):
LGTM

PR #427, opened 2022-11-21
Fixes latest docs validation link to grafana.com

cyriltovena:
Linter is happy but hard to validate those works.

@jdbaldry what do you think ?


simonswine (approved):
LGTM

PR #429, opened 2022-11-21
Ensure the doc-validator is using a particular tag

simonswine:
This will avoid breakages by more stricter versions of doc-validator.


cyriltovena (approved):
LGTM

PR #432, opened 2022-11-22
Enforce linux/amd64 platform for python

simonswine:
The python example only works on linux/amd64 or at least linux/arm64 is not supported. Also see https://github.com/timpalpant/mprofile/issues/3.


cyriltovena (approved):
LGTM

PR #415, opened 2022-11-09
Fixes local kubernetes cluster test

cyriltovena:
- Build correct binary arch and os.
- Remove resource requirement for running locally (otherwise it will lack memory)
- Fixes a typo in the README.md
- Replace datasource with the new Phlare datasource.

PR #434, opened 2022-11-23
Reduce maximal block duration for kubernetes cluster test

simonswine:
Based on #415 

This reduces the memory consumption of the ingesters


cyriltovena (commented):
Looks good !

PR #435, opened 2022-11-23
enhancement(helm): add note in resources section regarding max block duration

conorevans:
Signed-off-by: Conor Evans <coevans@tcd.ie>

Implements comment from grafana/pyroscope#2083

### DIFF
--- a/operations/phlare/helm/phlare/values.yaml
+++ b/operations/phlare/helm/phlare/values.yaml
@@ -66,7 +66,12 @@ phlare:
     # We usually recommend not to specify default resources and to leave this as a conscious
     # choice for the user. This also increases chances charts run on environments with little
     # resources, such as Minikube. If you do want to specify resources, uncomment the following
-    # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+    # lines, adjust them as necessary, and remove the curly braces after 'resources'.
+    #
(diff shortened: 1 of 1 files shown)


conorevans:
FYI @simonswine 👋

operations/phlare/helm/phlare/values.yaml
```diff
@@ -66,7 +66,12 @@ phlare:
     # We usually recommend not to specify default resources and to leave this as a conscious
     # choice for the user. This also increases chances charts run on environments with little
     # resources, such as Minikube. If you do want to specify resources, uncomment the following
-    # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+    # lines, adjust them as necessary, and remove the curly braces after 'resources'.
+    #
+    # Note that if memory consumption is higher than you would like, you can decrease the interval
+    # that profiles are written into blocks by setting `-phlaredb.max-block-duration` in the `extraArgs`
```
simonswine, line 72: I think a full example yaml would be even more helpful. What do you think if do it like this?
```suggestion
    # that profiles are written into blocks by setting `-phlaredb.max-block-duration` in the `extraArgs`
    
    extraArgs:
      phlaredb.max-block-duration: 30m
```

simonswine (approved):
Thanks for the contribution! Maybe take a look at my suggestion and let me know what you think.

PR #437, opened 2022-11-25
feat(objstore): support Tencent COS object storage

scottzhlin:
Tencent COS service is widely used in China, and currently, Thanos also supports COS as the storage backend of objstore.

So we can support COS as an object storage backend.


cyriltovena (approved):
LGTM

Issue #231, opened 2022-09-16
Reactivate log and metrics server instrumentation.


cyriltovena:
In https://github.com/grafana/fire/pull/226 we have started using bidi connect stream, however when logging (and may be metrics) middleware are activated (default), stream are getting stuck while communicating.

We should investigate and rewire.

PR #439, opened 2022-11-30
Fixes http2 instrumentation.

cyriltovena:
Since #226 we lost automatic http log and metrics. This PR re-introduce it by fixing the faulty middleware (log).

Fixes https://github.com/grafana/phlare/issues/231


simonswine (approved):
LGTM

Tried the metrics locally, they seem to work fine. Great work

PR #444, opened 2022-12-05
Stricter default for `podSecurityContext`

glindstedt:
Had to patch this in to conform with our own policies but seems like a good default to have


simonswine (approved):
LGTM!

Thanks for the contribution

PR #453, opened 2022-12-09
feat(agent/sd): add support HTTP service discovery

scottzhlin:
Add agent config to support HTTP service discovery, use Prometheus HTTP SD config.

### DIFF
--- a/docs/sources/operators-guide/configure/reference-configuration-parameters/index.md
+++ b/docs/sources/operators-guide/configure/reference-configuration-parameters/index.md
@@ -1246,6 +1246,10 @@ static_configs:
 kubernetes_sd_configs:
    [ - <kubernetes_sd_config> ... ]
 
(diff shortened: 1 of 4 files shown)


cyriltovena (approved):
LGTM

Merging this but ultimately we're moving toward Grafana Agent  see https://github.com/grafana/agent/pull/2552

scottzhlin:
> LGTM
> 
> Merging this but ultimately we're moving toward Grafana Agent see [grafana/agent#2552](https://github.com/grafana/agent/pull/2552)

Thanks @cyriltovena , I look forward to seeing the implementation and release of the Phlare Agent function on the Grafana Agent as soon as possible, and I will continue to pay attention. HTTP is a common service discovery mechanism, and it can be used on the lightweight Phlare Agent together with Kubenetes and static SD.

PR #454, opened 2022-12-09
Add docs for Django support

### LINKED ISSUE #447: Python Django integration
Hi team!

Wanting to add support for profiling our applications made using [Django](https://djangoproject.com/), I had to adapt a lot the provided Python example.

As I might not be first, nor the last person wanting to do this, I decided to extract what I did into a separate package that could be reused.

This package is called `django-pypprof`, and is available [on our GitLab organization](https://gitlab.com/prologin/tech/packages/django-pypprof), and it's also [mirrored on GitHub](https://github.com/prologin/django-pypprof). It's based a lot on [what pypprof does](https://github.com/timpalpant/pypprof/blob/master/pypprof/net_http.py) to expose profiles, but [adapted](https://gitlab.com/prologin/tech/packages/django-pypprof/-/blob/main/django_pypprof/views.py) for Django.

As such, I was wondering if it would be possible to link to this in a non-official way to allow people looking into how to do this to use this package.

Thanks!

rissson:
Closes #447 

Cc. @simonswine

### DIFF
--- a/docs/sources/operators-guide/configure-agent/language-support/python.md
+++ b/docs/sources/operators-guide/configure-agent/language-support/python.md
@@ -62,3 +62,70 @@ pprof -http :6060 "http://127.0.0.1:8081/debug/pprof/profile?seconds=5"
 ```
 
 [pprof]: https://github.com/google/pprof
+
+## How to instrument a Django application
+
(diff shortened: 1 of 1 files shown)


cyriltovena (approved):
Thanks a lot LGTM

PR #455, opened 2022-12-12
Default `service.port_name` to `http2`

glindstedt:
This ensures correct protocol selection in Istio and allows Phlare to work out of the box https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

Without this change queries break when using istio (something in the communication between querier and ingesters).


simonswine (approved):
LGTM!

Thanks for investigating this, that should be a fairly small impact change

cyriltovena (approved):

cyriltovena:
Not sure why the CI has not run here. I'll try a force push later

PR #459, opened 2022-12-19
Add gzip accept encoding

fredr:
Since the expected profile output should be gzip:ed.

This way the handler don't have to compress the output, but can instead rely on the webserver, if the implementor wants to.

The golang pprof cli sets this header as well.


cyriltovena (approved):
LGTM

PR #460, opened 2022-12-21
Update go.mod dependencies 

simonswine:
- Update vendor
- Update interface for copied group
- Refactor sorting to use upstream interface


cyriltovena (approved):
LGTM

Issue #463, opened 2022-12-29
Unable to set scrape timeout higher than scrape interval


eric-engberg:
#### Describe the bug
When attempting to set scrape interval to 10s and timeout to 30s the following error is given.
`failed creating phlare: scrape timeout must be larger or equal to inverval for: java`
when using
```
- job_name: "java"
  scrape_interval: "10s"
  scrape_timeout: "30s"
```

#### To Reproduce

1. Set timeout to a higher value than interval
2. Start phlare

#### Expected behavior

A higher timeout than interval should be accepted and work.

#### Environment

- Infrastructure: EKS
- Deployment tool: ArgoCD, Kustomize, Helm

cyriltovena:
I'll send a fix for that bug soon.

cyriltovena:
related to https://github.com/parca-dev/parca/pull/2220/files

PR #465, opened 2023-01-02
Fixes the scrape timeout validation.

cyriltovena:
Fixes https://github.com/grafana/phlare/issues/463

### DIFF
--- a/pkg/agent/profiles.go
+++ b/pkg/agent/profiles.go
@@ -228,7 +228,7 @@ func (tg *TargetGroup) targetsFromGroup(group *targetgroup.Group) ([]*Target, []
 				}
 
 				if pcfg, found := tg.config.ProfilingConfig.PprofConfig[profType]; found && pcfg.Delta {
-					params.Add("seconds", strconv.Itoa(int(time.Duration(tg.config.ScrapeInterval)/time.Second)))
+					params.Add("seconds", strconv.Itoa(int(time.Duration(tg.config.ScrapeInterval)/time.Second)-1))
 				}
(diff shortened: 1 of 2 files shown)


pkg/agent/profiles.go
```diff
@@ -228,7 +228,7 @@ func (tg *TargetGroup) targetsFromGroup(group *targetgroup.Group) ([]*Target, []
 				}
 
 				if pcfg, found := tg.config.ProfilingConfig.PprofConfig[profType]; found && pcfg.Delta {
-					params.Add("seconds", strconv.Itoa(int(time.Duration(tg.config.ScrapeTimeout)/time.Second)-1))
+					params.Add("seconds", strconv.Itoa(int(time.Duration(tg.config.ScrapeInterval)/time.Second)))
```
simonswine, line 231: I am not too sure why this had been removed. I always thought that this will make sure we have a consistent scrape time, even when network latency is added (up to 1 second).

Let's take this example:

scrape_interval: 15s
scrape_timeout: 18s
profiling_time: 15s (was 14s before) 

After this change, when we have network latency of 500ms, we would miss out on scrapes, because the real scrape interval would be 15s+network latency and this will create problems when we look at things over time e.g. this historgram screenshot (similar to the prometheus rate([twice_scrape_interval]) problem.

Here you can see a point missing at 11:12:47 because of that:


scrn-2023-01-11-11-13-41

simonswine (approved):
Good catch on the bug fix, not too sure if I agree on the second change.

PR #466, opened 2023-01-03
Add an API go module for external usage.

cyriltovena:
This adds a new go module  ( API ) that can be imported by other projects. It takes advantage of go workspace.

I had to remove `go mod verify` which is not yet compatible with workspace.

I've also taken the opportunity to reversion v1 to v1alpha1.

Once tagged `api/v0.1.1` the module should be available.

see https://go.dev/doc/modules/managing-source

This is a breaking change as path for each method have changed.


Makefile
```diff
@@ -108,7 +108,9 @@ go/lint: $(BIN)/golangci-lint
 .PHONY: go/mod
 go/mod:
 	GO111MODULE=on go mod download
-	GO111MODULE=on go mod verify
+	# doesn't work for go workspace
+	# GO111MODULE=on go mod verify
+	go work sync
 	GO111MODULE=on go mod tidy
```
simonswine, line 114: I think you also need to run `go mod tidy` in folder `api/`

simonswine (approved):
LGTM

Not an experent on go.work, but it looks good to me

Makefile
cyriltovena: I think so too.

PR #474, opened 2023-01-11
Implements a pprof query API.

cyriltovena:
The PR adds a new API endpoint for merging and querying profiles for a given label matcher and time range. It returns the results using the pprof format.


simonswine (approved):
This works for me, I think that will be super useful.

I have written #475 to test this properly and that seems to work well

PR #479, opened 2023-01-13
Fixes mapping slice insertion

cyriltovena:
In my previous PR I didn't realize but introduce a bug that would only surface when having many mapping.

It would use the id to index the slice to create, where it should just append to the slice.

PS: Tried to create a test but that requires more refactoring so I didn't bother, but happy to try to invest more time.


simonswine (approved):
LGTM

PR #481, opened 2023-01-16
Fix buckets range for distributor_received_samples.

cristiangreco:
This was incorrectly using a bitwise xor.


cyriltovena (approved):
LGTM

PR #484, opened 2023-01-17
Improve write path data management

simonswine:
* Ingest profiles into memory structure which is similar to parquet on-disk store
* Reuse query implementation for Head/Disk blocks a like

### TODO

* [ ] Remove old Head read implementation,
* [ ] Reuse buffers using `sync.Pool`s
* [ ] Refactor Head Flush to keep the current head
* [ ] Reuse block query implementation for new head


simonswine:
This will be superseeded by smaller PRs on subaspects, like #486

PR #486, opened 2023-01-20
Flush profiles to disk per row groups

simonswine:
This PR is flushes out profiles to a temporary on disk profile parquet file once a row group has been filled.

Further more it avoids using parquet.Buffer for sorting and rather uses a binary insert into the memory slice.

### DIFF
--- a/pkg/phlaredb/head_queriers.go
+++ b/pkg/phlaredb/head_queriers.go
@@ -144,7 +144,16 @@ func (q *headOnDiskQuerier) MergeByLabels(ctx context.Context, rows iter.Iterato
 }
 
 func (q *headOnDiskQuerier) Sort(in []Profile) []Profile {
-	// TODO: Use per row group order
+	var rowI, rowJ int64
+	sort.Slice(in, func(i, j int) bool {
+		if pI, ok := in[i].(BlockProfile); ok {
+			rowI = pI.RowNum
+		}
+		if pJ, ok := in[j].(BlockProfile); ok {
+			rowJ = pJ.RowNum
+		}
+		return rowI < rowJ
+	})
 	return in
 }
 
--- a/pkg/phlaredb/head_queriers.go
+++ b/pkg/phlaredb/head_queriers.go
@@ -23,48 +23,21 @@ type headOnDiskQuerier struct {
 	rowGroupIdx int
 }
 
+func (q *headOnDiskQuerier) rowGroup() *rowGroupOnDisk {
+	q.head.profiles.lock.RLock()
+	defer q.head.profiles.lock.RUnlock()
+	return q.head.profiles.rowGroups[q.rowGroupIdx]
+}
+
 func (q *headOnDiskQuerier) SelectMatchingProfiles(ctx context.Context, params *ingestv1.SelectProfilesRequest) (iter.Iterator[Profile], error) {
 	sp, ctx := opentracing.StartSpanFromContext(ctx, "SelectMatchingProfiles - HeadOnDisk")
 	defer sp.Finish()
 
-	index := q.head.profiles.index
-
-	ids, err := index.selectMatchingFPs(ctx, params)
+	// query the index for rows
+	rowIter, labelsPerFP, err := q.head.profiles.index.selectMatchingRowRanges(ctx, params, q.rowGroupIdx)
 	if err != nil {
 		return nil, err
 	}
-
-	index.mutex.RLock()
-
-	// gather rowRanges from matching series
-	var rowRanges = make(rowRanges, len(ids))
-	for _, fp := range ids {
-		// skip if series no longer in index
-		profileSeries, ok := index.profilesPerFP[fp]
-		if !ok {
-			continue
-		}
-
-		if q.rowGroupIdx >= len(profileSeries.profilesOnDisk) {
-			level.Error(q.head.logger).Log("msg", "unable to find matching rowGroup", "rowGroups", len(profileSeries.profilesOnDisk), "rowGroupIdx", q.rowGroupIdx)
-			continue
-		}
-
-		// skip if rowRange empty
-		rR := profileSeries.profilesOnDisk[q.rowGroupIdx]
-		if rR == nil {
-			continue
-		}
-
-		rowRanges[rR] = fp
-	}
-
-	// generate row number and fingerprint iter and release the index lock
-	var (
-		rowIter = rowRanges.fingerprintsWithRowNum()
-		rg      = q.head.profiles.rowGroups[q.rowGroupIdx]
-	)
-	index.mutex.RUnlock()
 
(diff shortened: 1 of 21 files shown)


simonswine:
With the latest changes I am also writing out the SeriesFingerprint to the parquet file. I need to do this to later renumber the SeriesIndexes. I think this also could help recover from an unclean shutdown.

I think this is ready for taking a first look @cyriltovena

pkg/phlaredb/profile_store.go
```diff
@@ -0,0 +1,452 @@
...
+		return nil
+	}
+
+	reader := parquet.NewGenericRowGroupReader[*schemav1.Profile](pg.rowGroup)
+	if err := reader.SeekToRow(pg.rowNum - 1); err != nil {
+		// TODO: Log error here
+		panic(fmt.Errorf("unable to seek row_num=%d from %s: %w", pg.rowNum, pg.rowGroup.file.Name(), err))
+	}
+
+	profiles := make([]*schemav1.Profile, 1)
```
cyriltovena, line 52: At scale, reading one row at a time will be problematic I think we should use flushed files to search for profiles instead.

I have some doubts about using profileGetter. Usually, you are querying for many profiles at once.

pkg/phlaredb/profile_store.go
```diff
@@ -0,0 +1,452 @@
...
+
+		// insert at the end of the slices
+		if len(s.slice) == posProfile {
+			s.slice = append(s.slice, p)
+			s.getters = append(s.getters, getters[pos])
+			continue
+		}
+
+		// make room at the correct position
+		s.slice = append(s.slice[:posProfile+1], s.slice[posProfile:]...)
```
cyriltovena, line 363: This can be quite heavy on the write path. I wonder why we need this in order when flushed files already provide this.

pkg/phlaredb/profile_store.go
simonswine: I guess then it probably would be best to keep the row numbers managed in the index, which contain a particular series. (rather than querying by row group).

pkg/phlaredb/profile_store.go
simonswine: We would need to go through a buffer to get the sorting as before. With only using the writer we would need to have it sorted before we pass it to the writer (when flushing a row group to disk).

In memory the sorting indeed doesn't provide any value. Alternatively we could just sort it once before flushing it to disk.

Let me compare those...

cyriltovena: Sorting once might be the way !

pkg/phlaredb/profile_store.go
```diff
@@ -0,0 +1,407 @@
...
+	fileCloser, err := s.prepareFile(path)
+	if err != nil {
+		return err
+	}
+	defer runutil.CloseWithErrCapture(&err, fileCloser, "closing row group segment file")
+
+	// order profiles properly
+	sort.Slice(s.slice, s.profileSort)
+
+	n, err := s.writer.Write(s.slice)
```
cyriltovena, line 211: Great !

This opens the opportunity to improve serialization later (by writing []parquet.Row)

pkg/phlaredb/profile_store.go
cyriltovena: Yes definitively.

cyriltovena:
I think for the read path we should narrow down column to what we need per type of request instead of exposing the whole profile object, the will help leverage parquet goodness by only pulling requested columns.


Request type are basically based on the querier interface

```go
type Querier interface {
...
	SelectMatchingProfiles(ctx context.Context, params *ingestv1.SelectProfilesRequest) (iter.Iterator[Profile], error)
	MergeByStacktraces(ctx context.Context, rows iter.Iterator[Profile]) (*ingestv1.MergeProfilesStacktracesResult, error)
	MergeByLabels(ctx context.Context, rows iter.Iterator[Profile], by ...string) ([]*typesv1.Series, error)
	MergePprof(ctx context.Context, rows iter.Iterator[Profile]) (*profile.Profile, error)
...
}
```

pkg/phlaredb/schemas/v1/profiles.go
```diff
@@ -25,6 +25,7 @@ var (
 	profilesSchema = parquet.NewSchema("Profile", phlareparquet.Group{
 		phlareparquet.NewGroupField("ID", parquet.UUID()),
 		phlareparquet.NewGroupField("SeriesIndex", parquet.Encoded(parquet.Uint(32), &parquet.DeltaBinaryPacked)),
+		phlareparquet.NewGroupField("SeriesFingerprint", parquet.Encoded(parquet.Uint(64), &parquet.DeltaBinaryPacked)),
```
cyriltovena, line 28: How do we use the stored SeriesFingerprint ?

simonswine:
@cyriltovena I would like to get another round of reviews, this afternoon I felt brave enough to deploy briefly to dev, but that resulted in all queries timing out, I will do look a bit more into this tomorrow

cyriltovena:
Looking good.

pkg/phlaredb/head_queriers.go
```diff
@@ -0,0 +1,309 @@
...
+	)
+
+	var (
+		profiles []Profile
+		buf      = make([][]parquet.Value, 1)
+	)
+	for pIt.Next() {
+		res := pIt.At()
+
+		v, ok := res.Entries[0].RowValue.(fingerprintWithRowNum)
```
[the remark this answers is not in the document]
simonswine: Yes exactly as the rows don't contain a fingerprint or series id and I am unsure which profiles will get filtered.

My thoughts this is the easiest way to get the fingerprint across. Also renamed the `rowNum()` method to make it clearer.

pkg/phlaredb/profile_store.go
```diff
@@ -0,0 +1,407 @@
...
+func (r *seriesIDRowsRewriter) ReadRows(rows []parquet.Row) (int, error) {
+	n, err := r.Rows.ReadRows(rows)
+	if err != nil {
+		return n, err
+	}
+
+	for pos, row := range rows[:n] {
+		// actual row num
+		rowNum := r.pos + int64(pos)
+		row[colIdxSeriesIndex] = parquet.ValueOf(r.seriesIndexes.getSeriesIndex(rowNum)).Level(0, 0, colIdxSeriesIndex)
```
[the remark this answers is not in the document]
simonswine: I have created #501 to track this idea

cyriltovena (approved):
LGTM

I think we resolve symbols multiple times for the same block, but that can be improved later.

simonswine:
> I think we resolve symbols multiple times for the same block, but that can be improved later.

I don't think it matters as long all of them are in memory

PR #487, opened 2023-01-24
Use gotestsum

simonswine:
Let me know what you think, but this makes the test output way more readable in CI.


cyriltovena (approved):
LGTM

PR #488, opened 2023-01-25
Update docker registry to point to docker hub.

simonswine:
Most likely a remnant from "before launch" times.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ GOPRIVATE=github.com/grafana/frostdb
 
 # Boiler plate for building Docker containers.
 # All this must go at top of file I'm afraid.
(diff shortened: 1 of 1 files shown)


cyriltovena (approved):
Will that push to dockerhub too in the CI ?

cyriltovena:
Wondering if this fixes https://github.com/grafana/pyroscope/issues/2109 ?

simonswine:
> Wondering if this fixes grafana/pyroscope#2109 ?

Good point, it can and maybe it should? What do you think?

> Will that push to dockerhub too in the CI ?

I think we might need some lightly different secrets for it, but in theory yes.

cyriltovena:
I want to push to dockerhub only I think so let's see

PR #490, opened 2023-01-26
Improve panic handling for the read path

### LINKED ISSUE #477: Protect the read path from crashing when panicking
It seems that if the read path is panicking, we will crash the whole pod. For ingester this can be very bad, we should recover and log them.

cyriltovena:
This ensures goroutines started on the read path are also recovering from panic if it ever happens, this way we don't bring down the write path.

Fixes #477


simonswine (approved):
LGTM. 

My concern is without a proper crash we might not notice those panics as much. How do you feel about a metrics (or maybe just a log rule) counting them up and an associated alert.

cyriltovena:
> LGTM.
> 
> My concern is without a proper crash we might not notice those panics as much. How do you feel about a metrics (or maybe just a log rule) counting them up and an associated alert.

The utility to recover panic does increment a counter. So we can add an alert for that.

cyriltovena:
I'll update the mixin :)

PR #491, opened 2023-01-27
Correctly implement connect health checker service

### LINKED ISSUE #472: GRPC healthcheck wrong port
#### Describe the bug
Seems like we register the healtch check endpoint using the GRPC port but then query from the client pool using the http port.
We should get rid of the GRPC port and verify it works correctly on http.

Ultimately we should find a way to remove the GRPC server since it's not required.

cyriltovena:
Connect healthcheck service is now correctly checking for services health and only registered on the http port.

Fixes #472

### DIFF
--- a/pkg/phlare/health.go
+++ b/pkg/phlare/health.go
@@ -1,0 +1,27 @@
+package phlare
+
+import (
+	"context"
+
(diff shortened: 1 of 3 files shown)


simonswine (approved):
LGTM

PR #495, opened 2023-02-02
Add observability using Loki/Tempo and Grafana Agent

simonswine:
TODOs:

* [x] Ensure the tempo datasource gets created
* [x] Add Loki for logs
* [ ] ~~Filter/don't sample /metrics /debug/pprof traces~~


tools/monitoring/environments/default/main.jsonnet
```diff
@@ -1 +1,2 @@
-import 'monitoring.libsonnet'
```
cyriltovena, line 1: This files is wiped on every deploy you should add to `monitoring.libsonnet` instead

cyriltovena (approved):
LGTM

PR #503, opened 2023-02-08
[release/v0.1.0] Update publishing workflows to use organization secret (#500)

jdbaldry:
The new tokens are managed centrally and have a longer expiry. Administrators of the grafanabot account will be
notified of the pending expiry and the secret can be rotated centrally without the need for a repository administrator to update their secrets.

The existing repository secrets can safely be removed. The tokens for those secrets will be removed by the end of this week.

Signed-off-by: Jack Baldry <jack.baldry@grafana.com>
(cherry picked from commit 3bdc62b24f0516eca4dc1b9026a5ef5e475af80c)

### DIFF
--- a/.github/workflows/publish-technical-documentation-next.yml
+++ b/.github/workflows/publish-technical-documentation-next.yml
@@ -25,7 +25,11 @@ jobs:
       uses: "actions/checkout@v3"
 
     - name: "Clone website-sync Action"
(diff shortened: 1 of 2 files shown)


simonswine:
I don't think this version will be maintained/updated, so l don't think it is worth backporting fixes

PR #506, opened 2023-02-08
Improve Worker timeout

cyriltovena:
This PR ensures the worker and frontend loop will automatically disconnect and reconnect before the handler timeout.

To ensure a query does not time out right after being processed in the querier we also increase the scheduler timeout by 2.

### DIFF
--- a/pkg/querier/worker/scheduler_processor.go
+++ b/pkg/querier/worker/scheduler_processor.go
@@ -49,12 +49,11 @@ var processorBackoffConfig = backoff.Config{
 
 func newSchedulerProcessor(cfg Config, handler RequestHandler, log log.Logger, reg prometheus.Registerer) (*schedulerProcessor, []services.Service) {
 	p := &schedulerProcessor{
-		log:            log,
-		handler:        handler,
-		maxMessageSize: cfg.GRPCClientConfig.MaxSendMsgSize,
-		querierID:      cfg.QuerierID,
-		grpcConfig:     cfg.GRPCClientConfig,
-		//
+		log:             log,
+		handler:         handler,
+		maxMessageSize:  cfg.GRPCClientConfig.MaxSendMsgSize,
+		querierID:       cfg.QuerierID,
+		grpcConfig:      cfg.GRPCClientConfig,
 		maxLoopDuration: cfg.MaxLoopDuration,
 
 		schedulerClientFactory: func(conn *grpc.ClientConn) schedulerpb.SchedulerForQuerierClient {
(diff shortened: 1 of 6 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -212,7 +212,9 @@ func (s *Scheduler) FrontendLoop(ctx context.Context, frontend *connect.BidiStre
 	if err != nil {
 		return err
 	}
-	defer s.frontendDisconnected(frontendAddress)
+	defer func() {
+		s.frontendDisconnected(frontendAddress)
+	}()
```
simonswine, line 217: Nit:  anonymous func()` likely from debugging, not really necessary anymore

pkg/scheduler/scheduler.go
```diff
@@ -425,7 +427,9 @@ func (s *Scheduler) QuerierLoop(ctx context.Context, bidi *connect.BidiStream[sc
 	querierID := resp.GetQuerierID()
 
 	s.requestQueue.RegisterQuerierConnection(querierID)
-	defer s.requestQueue.UnregisterQuerierConnection(querierID)
+	defer func() {
+		s.requestQueue.UnregisterQuerierConnection(querierID)
+	}()
```
simonswine, line 432: Nit:  anonymous func()` likely from debugging, not really necessary anymore

pkg/querier/worker/scheduler_processor.go
```diff
@@ -111,42 +119,67 @@ func (sp *schedulerProcessor) processQueriesOnSingleStream(workerCtx context.Con
...
+			select {
+			case <-timer.C:
+				level.Debug(sp.log).Log("msg", "waiting for inflight queries to complete")
+				for inflightQuery.Load() {
+					select {
+					case <-parentCtx.Done():
+						// In the meanwhile, the execution context has been explicitly canceled, so we should just terminate.
+						return
+					default:
+						time.Sleep(100 * time.Millisecond)
```
simonswine, line 163: You mind commenting why this is neccessary?

```suggestion
						time.Sleep(100 * time.Millisecond)
```

pkg/querier/worker/scheduler_processor.go
```diff
@@ -49,6 +54,8 @@ func newSchedulerProcessor(cfg Config, handler RequestHandler, log log.Logger, r
 		maxMessageSize: cfg.GRPCClientConfig.MaxSendMsgSize,
 		querierID:      cfg.QuerierID,
 		grpcConfig:     cfg.GRPCClientConfig,
+		//
```
simonswine, line 57: Remove or comment :slightly_smiling_face: 

```suggestion
		//
```

simonswine (approved):
LGTM

It is a bit fiddly to get it all working :slightly_smiling_face:. Let's see how many probelms we can see with this :slightly_smiling_face:

pkg/scheduler/scheduler.go
cyriltovena: true :)

PR #509, opened 2023-02-09
Fix regressions on the new ingress flush to disk handling.

simonswine:
- Fix adding new profiling series after row group segment has been cut.
- Close row Group file, before it gets reopened


cyriltovena (approved):
LGTM

PR #510, opened 2023-02-10
Add distributor limits

cyriltovena:
This PR introduce limits and runtime config with it to allow reloading limits per tenants.

For now only limits of ingestion are implemented but I'm planning to follow with more.

The overrides exporter is also added to export those limits as Prometheus metrics.

Usage:

`phlare -runtime-config.file=./overrides.yaml --config.file=./phlare.yaml`

```
cat overrides.yaml
overrides:
  "anonymous":
    ingestion_rate_mb: 0.1
    ingestion_burst_size_mb: 0.001
```

A lot of this code is adapted from Loki and Mimir.


pkg/validation/limits.go
```diff
@@ -0,0 +1,252 @@
...
+	// Distributor enforced limits.
+	IngestionRateMB        float64 `yaml:"ingestion_rate_mb" json:"ingestion_rate_mb"`
+	IngestionBurstSizeMB   float64 `yaml:"ingestion_burst_size_mb" json:"ingestion_burst_size_mb"`
+	MaxLabelNameLength     int     `yaml:"max_label_name_length" json:"max_label_name_length"`
+	MaxLabelValueLength    int     `yaml:"max_label_value_length" json:"max_label_value_length"`
+	MaxLabelNamesPerSeries int     `yaml:"max_label_names_per_series" json:"max_label_names_per_series"`
+
+	// Ingester enforced limits.
+	MaxLocalSeriesPerUser  int `yaml:"max_series_per_user" json:"max_series_per_user"`
+	MaxGlobalSeriesPerUser int `yaml:"max_global_series_per_user" json:"max_global_series_per_user"`
```
simonswine, line 31: I have spent a lot of time explaining the difference to GEx customers between global and local per instance limits. I wonder if we really need them both. What is the use case?

At least the yaml should be changing from `max_series_per_user` -> `max_local_series_per_user`

pkg/validation/limits.go
```diff
@@ -0,0 +1,252 @@
...
+
+	f.IntVar(&l.MaxLocalSeriesPerUser, "ingester.max-series-per-user", 0, "Maximum number of active series of profiles per user, per ingester. 0 to disable.")
+	f.IntVar(&l.MaxGlobalSeriesPerUser, "ingester.max-global-series-per-user", 5000, "Maximum number of active series of profiles per user, across the cluster. 0 to disable. When the global limit is enabled, each ingester is configured with a dynamic local limit based on the replication factor and the current number of healthy ingesters, and is kept updated whenever the number of ingesters change.")
+
+	_ = l.MaxQueryLength.Set("721h")
+	f.Var(&l.MaxQueryLength, "querier.max-query-length", "The limit to length of queries. 0 to disable.")
+
+	_ = l.MaxQueryLookback.Set("0s")
+	f.Var(&l.MaxQueryLookback, "querier.max-query-lookback", "Limit how far back in time series data and metadata can be queried, up until lookback duration ago. This limit is enforced in the query frontend. If the requested time range is outside the allowed range, the request will not fail, but will be modified to only query data within the allowed time range. The default value of 0 does not set a limit.")
```
simonswine, line 61: Replace time-series with profiling data

 ```suggestion
	f.Var(&l.MaxQueryLookback, "querier.max-query-lookback", "Limit how far back in profiling data can be queried, up until lookback duration ago. This limit is enforced in the query frontend. If the requested time range is outside the allowed range, the request will not fail, but will be modified to only query data within the allowed time range. The default value of 0 does not set a limit.")
```

File: pkg/phlare/modules.go [hunk 3]
```diff
@@ -104,14 +103,63 @@ func (f *Phlare) initQueryFrontend() (services.Service, error) {
...
+		// TenantLimits just delegates to RuntimeConfig and doesn't have any state or need to do
+		// anything in the start/stopping phase. Thus we can create it as part of runtime config
+		// setup without any service instance of its own.
+		f.TenantLimits = newTenantLimits(serv)
+	}
+
+	f.RuntimeConfig = serv
+
+	f.Server.HTTP.Methods("GET").Path("/runtime_config").Handler(runtimeConfigHandler(f.RuntimeConfig, f.Cfg.LimitsConfig))
+	f.Server.HTTP.Methods("GET").Path("/api/v1/user_limits").Handler(middleware.AuthenticateUser.Wrap(validation.UserLimitsHandler(f.Cfg.LimitsConfig, f.TenantLimits)))
```
simonswine, line 128: I really want to get rid of `user`  when we actually mean `tenant`.  I think now is the time

```suggestion
	f.Server.HTTP.Methods("GET").Path("/api/v1/tenant_limits").Handler(middleware.AuthenticateUser.Wrap(validation.UserLimitsHandler(f.Cfg.LimitsConfig, f.TenantLimits)))
```

simonswine (approved):
I really would like to consistently use `tenant` instead of `user`.

A couple of things to search/replace

* `userID`
* `PerUser`
* `per_user`

LGTM

pkg/validation/exporter/exporter.go
```diff
@@ -0,0 +1,204 @@
...
+	log log.Logger,
+	registerer prometheus.Registerer,
+) (*OverridesExporter, error) {
+	exporter := &OverridesExporter{
+		defaultLimits: defaultLimits,
+		tenantLimits:  tenantLimits,
+		overrideDescription: prometheus.NewDesc(
+			"phlare_limits_overrides",
+			"Resource limit overrides applied to tenants",
+			[]string{"limit_name", "user"},
```
simonswine, line 68:
```suggestion
			[]string{"limit_name", "tenant"},
```

pkg/validation/limits.go
cyriltovena: The use case is to set a max limit per ingester regardless of how many ingester exists in the ring. e.g small cluster.

cyriltovena:
Let's do another PR to use only `tenant`

PR #516, opened 2023-02-12
Use the correct API in readyHandler

hi-rustin:
I guess it's better to use the correct API to give an OK response. We should not use `http.Error` here.


cyriltovena (approved):
LGTM

PR #523, opened 2023-02-20
Ingester limits

cyriltovena:
This bring ingester limits:
- Global & Local maximum of active series
- Out of orders.

This is intentionally tracked outside of the storage package to not account for reset when data is flushed.

Builds on #510

### DIFF
--- a/pkg/phlaredb/profile_test.go
+++ b/pkg/phlaredb/profile_test.go
@@ -165,12 +165,12 @@ func Test_rowRangeIter(t *testing.T) {
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			it := rowRanges{tc.r: 0xff}.iter()
-			var result = []int64{}
+			result := []int64{}
 			for it.Next() {
 				result = append(result, it.At().RowNumber())
+				assert.Equal(t, model.Fingerprint(0xff), it.At().fp)
 			}
 			assert.Equal(t, tc.expected, result)
 		})
 	}
-
 }
--- a/pkg/phlaredb/profiles.go
+++ b/pkg/phlaredb/profiles.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sort"
 	"sync"
+	"time"
 	"unsafe"
 
 	"github.com/gogo/status"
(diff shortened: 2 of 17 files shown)


cyriltovena:
I still have to rename config yaml and flag.

pkg/phlaredb/profiles.go
```diff
@@ -136,6 +136,15 @@ type profileSeries struct {
...
 
+type PersistentProfileSeriesCallback interface {
+	// PreCreationSeries is called before creating a series to indicate if the series can be created.
+	// A non nil error means the series should not be created.
+	PreCreationSeries(labels.Labels) error
+	// PreCreationProfiles is called before creating profiles to indicate if the profile can be created.
+	// A non nil error means the profile should not be created.
+	PreCreationProfiles(ps *schemav1.Profile, lbs phlaremodel.Labels) error
+}
+
```
simonswine, line 147: Not too sure how this is to be used? Naming is not great super clear, maybe it PreProfileCreationCheck() and PreSeriesCreationCheck() is better, probably shouldn't be there until we use it

simonswine (approved):
See a few nits in the comments. LGTM

I have not actually tried to run it though. Also thanks for fixing the tenant/user thing.

pkg/phlaredb/profile_test.go
```diff
@@ -165,76 +165,12 @@ func Test_rowRangeIter(t *testing.T) {
...
 			for it.Next() {
 				result = append(result, it.At().RowNumber())
+				assert.Equal(t, model.Fingerprint(0xff), it.At().fp)
 			}
 			assert.Equal(t, tc.expected, result)
 		})
 	}
 }
-
-func Test_rowRangesIter(t *testing.T) {
```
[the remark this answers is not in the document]
cyriltovena: Sounds like a merge issue. Let me see.

pkg/phlaredb/profiles.go
cyriltovena: Yeah this is a left over. I'll remove it. Thanks.

PR #524, opened 2023-02-20
Upgrade linter to work with Go Generics

cyriltovena:
Update the linter to finally work with Generics. This has surface small interesting issue but also a lot of false negative with `unused` and generics.

But at least now they all run correctly.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -65,29 +65,20 @@ linters-settings:
 
 linters:
   enable:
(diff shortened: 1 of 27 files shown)


simonswine (approved):
LGTM. Thanks for going through them all and fixing it up :+1: 

Nice catch on the old grpc import path as well. Maybe we need faillint at some point for this

PR #529, opened 2023-02-21
Improve annotations based scraping for helm

cyriltovena:
This PR introduces annotations to configure scraping per profile type.

For example:

```yaml
metadata:
  annotations:
    profiles.grafana.com/memory.scrape: "true"
    profiles.grafana.com/memory.port: "8080"
    profiles.grafana.com/cpu.scrape: "true"
    profiles.grafana.com/cpu.port: "8080"
    profiles.grafana.com/goroutine.scrape: "true"
    profiles.grafana.com/goroutine.port: "8080"
```

will scrape the `memory`, `cpu` and `goroutine` profiles from the `8080` port of the pod.


cyriltovena:
The build is failing because we need to tag 0.5 with the new prometheus lib. But we can still merge this one.

simonswine (approved):
This seems to work for me. Thanks for getting this wall of YAML in :joy: 

LGTM

cyriltovena:
Thanks I'll merge and cut 0.5 :)

PR #531, opened 2023-02-21
Bump minimal support version to 1.19

hi-rustin:
1. Bump minimal support version to 1.19
If we use 1.18 to `make lint`, then we will get an error:
```sh
# honnef.co/go/tools/unused
../../go/pkg/mod/honnef.co/go/tools@v0.4.2/unused/unused.go:419:14: obj.Origin undefined (type *types.Var has no field or method Origin)
../../go/pkg/mod/honnef.co/go/tools@v0.4.2/unused/unused.go:421:14: obj.Origin undefined (type *types.Func has no field or method Origin)
note: module requires Go 1.19
# honnef.co/go/tools/staticcheck
../../go/pkg/mod/honnef.co/go/tools@v0.4.2/staticcheck/lint.go:4025:29: righti.Method(i).Origin undefined (type *types.Func has no field or method Origin)
../../go/pkg/mod/honnef.co/go/tools@v0.4.2/staticcheck/lint.go:4030:36: sel.Obj().(*types.Func).Origin undefined (type *types.Func has no field or method Origin)
note: module requires Go 1.19
```
2. Increase the timeout to 10m, because I found it failed twice.

PR #532, opened 2023-02-21
Increase linter timeout from 5 to 10m

cyriltovena:


cyriltovena:
Already covered by #531

PR #534, opened 2023-02-21
Move from kubeval to kubeconform

### LINKED ISSUE #520: Replace kube-eval with kube-conform
We kube-eval for testing the generated helm charts. It is no longer maintained and has some issues on darwin/arm64.

We should consider switching to https://github.com/yannh/kubeconform

simonswine:
Fixes #520


cyriltovena (approved):
LGT M1 ;)

PR #535, opened 2023-02-22
update helm version to 0.5.0

cyriltovena:
to merge once 0.5 is live.


simonswine (commented):
This needs to go into `main` instead.

There also seems to be a bug with #497 if I take a look at the helm chart test. I will try to find it.

cyriltovena:
it was but let's skip 0.5 for helm.

PR #541, opened 2023-02-23
Remove addresses on symbolized profiles

cyriltovena:
We realised that this create a lot of different stacktraces while just the address changes but not the binary itself. 

In the future we'll introduce buildid to recognized binary change, so when a profiles is symbolized we can just get rids on the adresses.


simonswine (approved):
Thanks for looking into this. LGTM

cyriltovena:
Waiting to run this for a day and I'll merge

cyriltovena:
Going to push this to next week, I still want to investigate some block.

PR #542, opened 2023-02-23
Support set test mode for flags test

hi-rustin:
See https://github.com/grafana/phlare/pull/497#issuecomment-1438260905

### DIFF
--- a/cmd/phlare/main.go
+++ b/cmd/phlare/main.go
@@ -16,10 +16,6 @@ import (
 	_ "github.com/grafana/phlare/pkg/util/build"
 )
 
-// testMode is used to test flag parsing.
-// We don't want to exit the process when running tests.
-var testMode = false
-
 type mainFlags struct {
 	phlare.Config `yaml:",inline"`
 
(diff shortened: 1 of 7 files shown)


cmd/phlare/main.go
```diff
@@ -49,14 +53,20 @@ func main() {
 		flags mainFlags
 	)
 
-	if err := cfg.DynamicUnmarshal(&flags, os.Args[1:], flag.CommandLine); err != nil {
+	if err := cfg.DynamicUnmarshal(&flags, os.Args[1:], flag.CommandLine, testMode); err != nil {
```
hi-rustin, line 56: I am not sure does this is a good way. Maybe it is too ugly. If you have any suggestions please feel free to comment. Thanks!

cmd/phlare/main.go
```diff
@@ -16,6 +16,10 @@ import (
 	_ "github.com/grafana/phlare/pkg/util/build"
 )
 
+// testMode is used to test flag parsing.
+// We don't want to exit the process when running tests.
+var testMode = false
```
simonswine, line 21: I would move this as a private global to `pkg/cfg/dynamic.go`.

Then I would have a public method `SetTestMode(bool)` in `pkg/cfg/dynamic_test.go`. This doesn't pollute the arguments. How do you feel about that?

cmd/phlare/main.go
simonswine: https://github.com/grafana/phlare/pull/542/files#r1116866328

cmd/phlare/main.go
hi-rustin: Yeah. It sounds good! Thanks for your suggestion!

hi-rustin (commented):
Self check

simonswine (approved):
Thanks for addressing this case :+1: 

LGTM

PR #544, opened 2023-02-24
Fix deprecated set-output warnings and bump actions/cache to v3

hi-rustin:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


simonswine (approved):
LGTM

Issue #494, opened 2023-02-02
Mixin: Make the UID of the "Reads" and "Writes" Dashboards more unique


SeamusGrafana:
**Is your feature request related to a problem? Please describe.**

The UID's used by the "Reads" and "Writes" Dashboards are just `reads` and `writes` and is not very unique, as there are other dashboards that use that same UID (for example, the Loki Mixin).

Examples from Phlare Mixin:

* https://github.com/grafana/phlare/blob/main/operations/phlare/jsonnet/phlare-mixin/phlare-mixin/dashboards/phlare-reads.libsonnet#L31
* https://github.com/grafana/phlare/blob/main/operations/phlare/jsonnet/phlare-mixin/phlare-mixin/dashboards/phlare-writes.libsonnet#L34

Examples from Loki Mixin:

* https://github.com/grafana/loki/blob/main/production/loki-mixin/dashboards/loki-reads.libsonnet#L44
* [<https://github.com/grafana/loki/blob/main/production/loki-mixin/dashboards/loki-writes.libsonnet>](https://github.com/grafana/loki/blob/main/production/loki-mixin/dashboards/loki-writes.libsonnet#L37)

**Describe the solution you'd like**

I would like to see the UID become _more_ unique, e.g. `phalre-reads` or `phlare-writes`, or even a UUID or similar.

**Describe alternatives you've considered**

While I can rename the UIDs while importing Dashboards, the idea should be they are unique and so this should not be needed. This also makes automated provisioning more complex as there are more collisions.

**Additional context**

A similar request has also be opened in the Loki repo ([here](https://github.com/grafana/loki/issues/8386)), to make those more `unique` as well.

cyriltovena:
Sounds good  ! feel free to send a PR to solve it.

hi-rustin:
@cyriltovena I guess we can close this issue now.

PR #545, opened 2023-02-26
Mixin: Make the UID of the "Reads" and "Writes" Dashboards more unique

SeamusGrafana:
Related to #494 

Make the UID of the `Reads` and `Writes` Mixin Dashboards more unique by appending `phlare-` to them, e.g. `writes` becomes `phlare-writes`, this avoid conflicts with other Mixins (such as the Loki one).


simonswine (approved):
Thanks for this change. LGTM

I think is probably also worth changing this on the loki end

PR #546, opened 2023-02-27
Rename `Dockerfile.debug` to `debug.Dockerfile`

hi-rustin:
I think it is better to use `debug.Dockerfile` instead of `Dockerfile.debug`. Because if we use `debug` as the postfix, then IDEs can not recognize it as a docker file. It is difficult to edit it in GoLand.(You need to config your GoLand to recognize it as a docker file manually.) So maybe using `Dockerfile` as the postfix is better.


simonswine (commented):
Not opposed to this rename, but you still need to change the Makefile, as `make docker-image/phlare/build-debug` would fail after this rename:

```
$ make docker-image/phlare/build-debug
[...]
ERROR: failed to solve: failed to read dockerfile: open /tmp/buildkit-mount2538699037/Dockerfile.debug: no such file or directory
make: *** [Makefile:160: docker-image/phlare/build-debug] Error 1
```

hi-rustin:
> Not opposed to this rename, but you still need to change the Makefile, as `make docker-image/phlare/build-debug` would fail after this rename:

Thanks for your review!  💚 💙 💜 💛 ❤️ Update!

PR #550, opened 2023-02-27
Validates that push request contains at least one sample

### LINKED ISSUE #543: Distributor hangs with empty push requests
Sounds like the distributor will hangs if the push request is empty. It seems to wait on a channel but no goroutine is fired up.

We should probably validate request and return 400.

cyriltovena:
Fixes #543


simonswine (approved):
LGTM

PR #551, opened 2023-02-28
Adds tenant overrides in helm

cyriltovena:


operations/phlare/helm/phlare/values.yaml
```diff
@@ -710,6 +710,12 @@ phlare:
             regex: Pending|Succeeded|Failed|Completed
             action: drop
 
+  # Allows to add tenant specific overrides to the default limit configuration.
```
simonswine, line 713: If you do this, it should show up in the README.md correctly

```suggestion
  # -- Allows to add tenant specific overrides to the default limit configuration.
```

simonswine (approved):
LGTM

PR #552, opened 2023-03-01
Bump checkout@v2 to checkout@v3

hi-rustin:
Some actions are using checkout@v2 and some are using checkout@v3. Just bump all checkout@v2 to checkout@v3.


cyriltovena (approved):
LGTM

PR #554, opened 2023-03-02
Fixed crashed rust example on M1 MacOS

hi-rustin:
I tried to run the rust example on my M1 Macbook Pro. When I try to get the profile, it crashed. I think this is an issue from `pprof-rs`.

So I just fix it by using [Platform specific dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies)

See more at: https://github.com/tikv/pprof-rs/issues/131


cyriltovena (approved):
LGTM

PR #555, opened 2023-03-02
Use IEC size format for rate_limit errors

cyriltovena:
This is what we use in the overrides file format and will avoid confusion in the future.


simonswine (approved):
LGTM

PR #559, opened 2023-03-06
Fixes no profiles to merge error

### LINKED ISSUE #557: Error when no profile found to merge
Seems like if ingester have no profiles for the request we error but we should probably not:

```
unknown: no profiles to merge
```

cyriltovena:
Fixes grafana/phlare#557 

This was happening when the timerange was returning no queriers.

Also changed the TimeNanos to be from the end not the start.

### DIFF
--- a/pkg/model/profile.go
+++ b/pkg/model/profile.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"github.com/gogo/status"
+	"github.com/google/pprof/profile"
 	"github.com/prometheus/prometheus/model/labels"
 	"google.golang.org/grpc/codes"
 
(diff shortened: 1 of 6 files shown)


pkg/model/profile.go
```diff
@@ -45,3 +46,18 @@ func SelectorFromProfileType(profileType *typesv1.ProfileType) *labels.Matcher {
...
 	}
 }
+
+// SetProfileMetadata sets the metadata on the profile.
+func SetProfileMetadata(p *profile.Profile, ty *typesv1.ProfileType) {
+	p.SampleType = []*profile.ValueType{{Type: ty.SampleType, Unit: ty.SampleUnit}}
+	p.DefaultSampleType = ty.SampleType
+	p.PeriodType = &profile.ValueType{Type: ty.PeriodType, Unit: ty.PeriodUnit}
+	switch ty.Name {
+	case "process_cpu": // todo: this should support other types of cpu profiles
```
simonswine, line 56: Should we create an issue for that?

simonswine (approved):
LGTM

pkg/model/profile.go
cyriltovena: I already did. https://github.com/grafana/phlare/issues/560

PR #573, opened 2023-03-09
Ignore weekly release tags (2nd try)

simonswine:


cyriltovena (commented):
LGTM

cyriltovena (approved):

simonswine:
> you may only define one of `tags` and `tags-ignore` for a single event

PR #574, opened 2023-03-09
Refactor metrics in PhlareDB

cyriltovena:
Not only this is easier to manage, but this is also more resilient.

### DIFF
--- a/pkg/util/prometheus.go
+++ b/pkg/util/prometheus.go
@@ -1,0 +1,23 @@
+package util
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// RegisterOrGet registers the collector c with the provided registerer.
+// If the registerer is nil, the collector is returned without registration.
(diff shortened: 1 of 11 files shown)


pkg/util/prometheus.go
```diff
@@ -0,0 +1,23 @@
...
+func RegisterOrGet[T prometheus.Collector](reg prometheus.Registerer, c T) T {
+	if reg == nil {
+		return c
+	}
+	err := reg.Register(c)
+	if err != nil {
+		already, ok := err.(prometheus.AlreadyRegisteredError)
+		if ok {
+			return already.ExistingCollector.(T)
+		}
```
simonswine, line 19: Nice, I wasn't aware that you have that custom error type

simonswine (approved):
Thanks for cleaning that up!

Tried it locally works well :+1: 

LGTM

PR #576, opened 2023-03-10
Use goreleaser to build weekly releases

simonswine:
Those releases power the Grafana Labs internal roll-outs, that occur
weekly into our production systems.

Goreleaser is used to build those releases with the same config that is
used for the project releases.

To get around some limitations around semver in goreleaser we have to
imitate a version with local git tags.


cyriltovena (approved):
LGTM

PR #579, opened 2023-03-13
Use date instead of timestamp

simonswine:
A change that I made for the weekly releases using goreleases #576  was display the unixtime instead.


cyriltovena (approved):
LGTM

PR #580, opened 2023-03-14
Make bucket range test more maintainable

hi-rustin:
- remove useless convert
- use some consts to make bucket range test more maintainable


simonswine (approved):
Thanks that makes it a lot more readable.

LGTM

PR #583, opened 2023-03-14
Add instructions on how to join the Grafana Slack workspace

prati0100:
When one clicks the channel link Slack asks for an email to log into the Grafana Slack workspace. That email needs to be invited to the workspace first. It is not obvious how to do that. Add a link to slack.grafana.com which can be used to get the invitation.


simonswine (approved):
Very good point! Thanks for adding this

PR #584, opened 2023-03-14
Add new metrics for samples bytes ingestion without symbols

cyriltovena:


simonswine (approved):
LGTM

Is it worth also measuring the size of symbols?

PR #585, opened 2023-03-15
Add the home page link for Phlare logo

hi-rustin:
Add a home page link for the Phlare logo. This would help us go back to the home page.
Also, it is quite common on other websites.


cyriltovena (approved):
LGTM

PR #594, opened 2023-03-20
Improve profiles flushing using ReadRowsFrom

cyriltovena:
This adds a flush benchmark and an optimization to reduce memory usage.

```
name      old time/op    new time/op    delta
Flush-16     400ms ± 8%     249ms ±10%  -37.76%  (p=0.008 n=5+5)

name      old alloc/op   new alloc/op   delta
Flush-16    1.03GB ± 0%    0.20GB ± 1%  -80.12%  (p=0.008 n=5+5)

name      old allocs/op  new allocs/op  delta
Flush-16     8.26k ± 1%     3.43k ± 1%  -58.54%  (p=0.008 n=5+5)
```

### DIFF
--- a/pkg/phlaredb/profile_store.go
+++ b/pkg/phlaredb/profile_store.go
@@ -275,8 +275,13 @@ func (s *profileStore) writeRowGroups(path string, rowGroups []parquet.RowGroup)
 		if err != nil {
 			return 0, 0, err
 		}
+
 		n += uint64(nInt64)
 		numRowGroups += 1
+
+		if err := s.writer.Flush(); err != nil {
+			return 0, 0, err
+		}
 	}
 
 	if err := s.writer.Close(); err != nil {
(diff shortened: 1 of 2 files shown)


simonswine:
We also need to call `Flush()` to ensure the rowGroup boundaries are still respected.

So that one segment in becomes one row group in the final file. Otherwise the ordering is incorrect. Still significant win:

```
❯ benchstat before.txt after2.txt
name     old time/op    new time/op    delta
Flush-8     533ms ±13%     294ms ± 6%  -44.84%  (p=0.008 n=5+5)

name     old alloc/op   new alloc/op   delta
Flush-8    1.03GB ± 0%    0.20GB ± 1%  -80.64%  (p=0.016 n=4+5)

name     old allocs/op  new allocs/op  delta
Flush-8     8.00k ± 0%     5.07k ± 1%  -36.62%  (p=0.016 n=4+5)
```

simonswine (approved):
Your bit LGTM

PR #595, opened 2023-03-22
Allow environment variables from secrets/configMaps

simonswine:
This allows Helm values to be complex and hence specify
configMap/secrets as valueFrom soruces for environment variables.


cyriltovena (approved):
LGTM

PR #596, opened 2023-03-22
feat: add pyroscope ingestion endpoint

korniltsev:
This is a poc introducing pyroscope's /ingest api for push mode.
It reuses pyrscope OSS's ingestionHandler and parsers, this allows to ingest all supported formats, like jfr, etc.
It is not supper efficient at the moment, for example for a pprof profile there are a number of conversions happening: 'serialized pprof' - 'pprof' - 'pyroscope trees' - 'pprof' - 'serialized pprof' and only then passed to the phlare ingestion service.

### DIFF
--- a/pkg/phlare/modules.go
+++ b/pkg/phlare/modules.go
@@ -370,8 +370,7 @@ func (f *Phlare) initIngester() (_ services.Service, err error) {
 		return nil, err
 	}
 	ingesterv1connect.RegisterIngesterServiceHandler(f.Server.HTTP, svc, f.auth)
-	//pyroscopePath := "/pyroscope/ingest"
-	pyroscopePath := "/ingest"
+	pyroscopePath := "/pyroscope/ingest"
 	f.Server.HTTP.Handle(pyroscopePath, util.AuthenticateUser(f.Cfg.MultitenancyEnabled).Wrap(pyroscope.NewPyroscopeIngestHandler(svc, f.logger)))
 	return svc, nil
 }


pkg/ingester/pyroscope/ingest_adapter.go
```diff
@@ -0,0 +1,141 @@
...
+			continue
+		}
+		series.Labels = append(series.Labels, &typesv1.LabelPair{
+			Name:  k,
+			Value: v,
+		})
+	}
+	series.Samples = []*pushv1.RawSample{{
+		RawProfile: b,
+		ID:         uuid.New().String(), //todo generate on a client
```
cyriltovena, line 81: it's fine to have it here. IMO

pkg/ingester/pyroscope/ingest_adapter.go
```diff
@@ -0,0 +1,213 @@
...
+	})
+	newString := func(value string) int64 {
+		id := int64(len(pprof.StringTable))
+		pprof.StringTable = append(pprof.StringTable, value)
+		return id
+	}
+	// todo(korniltsev) the pprof is partially built by pyroscope oss and partially here, do it in one place
+	pprof.Mapping = []*tree.Mapping{{Id: 0}} // a fake mapping
+	if pi.SampleRate != 0 && (metric == metricWall || metric == metricProcessCPU) {
+		pprof.Period = 1_000_000_000 / int64(pi.SampleRate)
```
petethepig, line 75: it's not obvious what `1_000_000_000` is

petethepig:
Talked to @korniltsev — he's going to be finishing this PR today.

pkg/phlare/modules.go
```diff
@@ -364,12 +365,15 @@ func (f *Phlare) context() context.Context {
...
 
-	ingester, err := ingester.New(f.context(), f.Cfg.Ingester, f.Cfg.PhlareDB, f.storageBucket, f.Overrides)
+	svc, err := ingester.New(f.context(), f.Cfg.Ingester, f.Cfg.PhlareDB, f.storageBucket, f.Overrides)
 	if err != nil {
 		return nil, err
 	}
-	ingesterv1connect.RegisterIngesterServiceHandler(f.Server.HTTP, ingester, f.auth)
-	return ingester, nil
+	ingesterv1connect.RegisterIngesterServiceHandler(f.Server.HTTP, svc, f.auth)
+	//pyroscopePath := "/pyroscope/ingest"
```
petethepig, line 373: @cyriltovena @simonswine Not sure how you all feel about the prefix? I know /render and /label-values [have the prefix](https://github.com/grafana/phlare/pull/591/files#diff-ff08860a084147a57eed45febf9519697bc6179c98981010040b66ffde458e79R208-R209)

I don't feel strongly one way or the other.

cyriltovena: Fine for me

pkg/phlaredb/delta.go
```diff
@@ -97,12 +100,21 @@ func copySample(s *schemav1.Sample) *schemav1.Sample {
 }
 
 func isDelta(lbs phlaremodel.Labels) bool {
+	if lbs.Get(phlaremodel.LabelNameDelta) == "false" {
```
cyriltovena, line 103: Nit

It's a bit confusing that you can only use the label to deactivate.

May be if the labels is true then we should return true too and if not a Boolean like empty go to the rest of the function.

pkg/phlare/modules.go
petethepig: FWIW, I added `/pyroscope` prefix [here](https://github.com/grafana/backend-enterprise/pull/4418/files#diff-ef48fc318181e7feb0e84c93de69f745eec72eb9450764e97a15866530656c88R370) for consistency, so I guess now I do lean slightly more towards having the prefix.

cyriltovena (approved):
Lgtm

pkg/phlare/modules.go
```diff
@@ -364,12 +365,15 @@ func (f *Phlare) context() context.Context {
...
+	pyroscopePath := "/ingest"
```
petethepig, line 374:
```suggestion
	pyroscopePath := "/pyroscope/ingest"
```

PR #602, opened 2023-03-29
Update to setup-go@v4 github action

cyriltovena:
hoping caching will works in github actions.

Seems like we are go downloading everything, everytime.


simonswine (approved):
LGTM

In theory this line should only throw the cache away when `go.mod` changes:

https://github.com/grafana/phlare/blob/7b8032dd972999c47a35985ca59b368b33c7b7a8/.github/workflows/test.yml#L41

PR #604, opened 2023-03-30
fix: allow __delta__ label to force-enable as well as force-disable delta computation

korniltsev:
https://github.com/grafana/phlare/pull/596#discussion_r1151055486


cyriltovena (approved):
LGTM

PR #606, opened 2023-03-31
Serve a basic react app under /ui

eh-am:
This is an initial step to get a subset of the Pyroscope OSS UI in Phlare. 

In this PR, a simple React App will be build and served under `/ui` when the single binary is run.
When developing locally, it will be served under its own port, in a decoupled fashion. This is so that the frontend can be developed without backend serving the frontend, which is faster and easier.

It leads to a distinction between "prod/dev builds", where in "prod" it adds the `embedassets` build flag, which expects the `public/build` directory to exist. By default, `make build` became this prod built.

For dev builds (builds that don't require the frontend to have been built), there's a `build-dev` command.


.goreleaser.yaml
```diff
@@ -22,6 +22,7 @@ builds:
       - -trimpath
     tags:
       - netgo
+      - embedassets
```
cyriltovena, line 25: I guess we need to build frontend too before go binaries ?

eh-am: https://github.com/grafana/phlare/pull/606/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R48 it's here i believe

cyriltovena: ho I see. yep

cyriltovena:
> be at som

Yeah I think we will have to !

PR #613, opened 2023-04-04
Improve memory ownership at pprof parsing

kolesnikovae:
In some cases an object allocated from the pool is not returned back to the pool. The change is aimed at improving ownership of the objects allocated in pools at pprof parsing specifically, which affects both distributor and ingester services

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -15,6 +15,7 @@ import (
 	"github.com/grafana/dskit/services"
 	"github.com/prometheus/client_golang/prometheus"
 
+	profilev1 "github.com/grafana/phlare/api/gen/proto/go/google/v1"
 	ingesterv1 "github.com/grafana/phlare/api/gen/proto/go/ingester/v1"
 	pushv1 "github.com/grafana/phlare/api/gen/proto/go/push/v1"
 	phlareobjstore "github.com/grafana/phlare/pkg/objstore"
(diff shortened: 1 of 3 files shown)


pkg/ingester/ingester.go
```diff
@@ -188,29 +189,28 @@ func (i *Ingester) Push(ctx context.Context, req *connect.Request[pushv1.PushReq
...
+							validation.DiscardedProfiles.WithLabelValues(string(reason), instance.tenantID).Add(float64(1))
+							validation.DiscardedBytes.WithLabelValues(string(reason), instance.tenantID).Add(float64(size))
+							return err
 						}
 					}
+					return nil
+				})
+				if err != nil {
 					return nil, err
 				}
```
kolesnikovae, line 213: I'm not sure if we need to break the loop on the first error encountered. If we do need, then I guess we should bump the `DiscardedProfiles` metrics by the number of profiles we actually dropped, like `len(series.Samples)-i`. However, it does not work for `DiscardedBytes`

cyriltovena (approved):
LGTM

pkg/ingester/ingester.go
simonswine: Yes I fully agree, we should ingest all the profiles that are valid (== soft error), and then returning the error appropriate for the offending sample. Otherwise we would loose perfectly valid samples. 

This is how mimir is doing it: https://github.com/grafana/mimir/blob/main/pkg/ingester/ingester.go#L826-L828

pkg/ingester/ingester.go
```diff
@@ -188,29 +189,28 @@ func (i *Ingester) Push(ctx context.Context, req *connect.Request[pushv1.PushReq
...
 						case validation.OutOfOrder:
-							return nil, connect.NewError(connect.CodeInvalidArgument, err)
+							return connect.NewError(connect.CodeInvalidArgument, err)
 						case validation.SeriesLimit:
-							return nil, connect.NewError(connect.CodeResourceExhausted, err)
+							return connect.NewError(connect.CodeResourceExhausted, err)
+						default:
+							validation.DiscardedProfiles.WithLabelValues(string(reason), instance.tenantID).Add(float64(1))
+							validation.DiscardedBytes.WithLabelValues(string(reason), instance.tenantID).Add(float64(size))
+							return err
```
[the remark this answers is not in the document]
simonswine: I am not too sure what currently would be an unknown reason, but I think that makes sense and we would know we need to add a new reason.

pkg/ingester/ingester.go
simonswine: I agree the scope of that is beyond this PR.

simonswine (approved):
LGTM

PR #619, opened 2023-04-06
Use setup-go's caching

simonswine:
This changes uses setup-go's cache and disables the manual approach by
the separate steps.

In order to make setup-go's cache work we have to do the checkout before
the setup-go.


cyriltovena (approved):
LGTM

PR #620, opened 2023-04-11
feat: make tags work in the ui

eh-am:
https://user-images.githubusercontent.com/6951209/231123405-8717515b-61f9-42b3-a0a2-0571220842eb.mov

---
Adds supports to the labels/tags selector in the UI. It uses the `querier.v1.QuerierService/Series` API as [suggested by @simonswine](https://github.com/grafana/phlare/pull/620#pullrequestreview-1380893609)

* It's all done client side. The server returns way more data than needed, but decided to take the simple approach for now and fix in the frontend.
* It only shows tags/labels that are valid for that `profileTypeId`. For example, it doesn't show dynamic tags for memory, only CPU.

### DIFF
--- a/pkg/phlare/modules.go
+++ b/pkg/phlare/modules.go
@@ -211,6 +211,7 @@ func (f *Phlare) initQuerier() (services.Service, error) {
 
 	f.Server.HTTP.Handle("/pyroscope/render", util.AuthenticateUser(f.Cfg.MultitenancyEnabled).Wrap(http.HandlerFunc(querierSvc.RenderHandler)))
 	f.Server.HTTP.Handle("/pyroscope/label-values", util.AuthenticateUser(f.Cfg.MultitenancyEnabled).Wrap(http.HandlerFunc(querierSvc.LabelValuesHandler)))
+	f.Server.HTTP.Handle("/pyroscope/labels", util.AuthenticateUser(f.Cfg.MultitenancyEnabled).Wrap(http.HandlerFunc(querierSvc.LabelNamesHandler)))
 
 	sm, err := services.NewManager(querierSvc, worker)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


simonswine (commented):
I think this already goes in a good way.

I think we should at least support matchers (as they are supported by our index as present (as opposed to the exact timestamps supplied by from/to)

Here is a curl example of this:

```
curl -sL -d '{"matchers": ["{__profile_type__=\"goroutine:goroutine:count:goroutine:count\"}"]}' -X POST -H "content-type: application/json" https://play-phlare.grafana.org/querier.v1.QuerierService/Series
{"labelsSet":[{"labels":[{"name":"__name__", "value":"goroutine"}, {"name":"__period_type__", "value":"goroutine"}, {"name":"__period_unit__", "value":"count"}, {"name":"__profile_type__", "value":"goroutine:goroutine:count:goroutine:count"}, {"name":"__type__", "value":"goroutine"}, {"name":"__unit__", "value":"count"}, {"name":"app_kubernetes_io_component", "value":"all"}, {"name":"app_kubernetes_io_instance", "value":"phlare"}, {"name":"app_kubernetes_io_name", "value":"phlare"}, {"name":"container", "value":"phlare"}, {"name":"controller_revision_hash", "value":"phlare-58775689d6"}, {"name":"instance", "value":"10.145.176.117:4100"}, {"name":"job", "value":"kubernetes-pods-name"}, {"name":"name", "value":"phlare"}, {"name":"namespace", "value":"profiles-prod-play"}, {"name":"pod", "value":"phlare-0"}, {"name":"statefulset_kubernetes_io_pod_name", "value":"phlare-0"}]}]
```

eh-am:
Actually still have to handle the default case, when the backend doesn't return anything.

simonswine (approved):
LGTM.

This works well for the single binary mode. But we I try the micro-services mode, it misses the pyroscope endpoints. (I will look into a separate PR).

PR #622, opened 2023-04-12
feat: add diff support

eh-am:
Add support for diff:
Screenshot 2023-04-24 at 14 33 54

Given 2 trees, it generates a `*querierv1.FlameGraphDiff` in diff format. I preferred this way (instead of reusing `Flamegraph`) to make it clear its semantics and shape are the same (for example, `leftTicks/rightTicks` are required).

Before implementing I had 2 approaches in mind:
* Write code to convert from `phlare's tree` into `pyroscope's tree`, then just defer to pyroscope's code.
* Copy the [diff implementation from pyroscope OG](https://github.com/grafana/pyroscope/blob/main/pkg/storage/tree/treediff.go).

Ended up using the second approach, since it doesn't require an extra conversion, plus it gives us more flexibility.

Added the bare minimum changes to get it working. Haven't checked performance nor anything like that. So feel free to suggest improvements :)

# Drawbacks
* It only supports positive nodes. I added an assertion at the beginning. Don't fully understand the data model, so decided to fail early and deal with it properly when we have real data that fails this assertion.


Rperry2174 (approved):
LGTM -- I'd say just create an issue for the outstanding items so we don't forget, but I think we can merge this and follow up in future PRs. 

Nice job!!! 

image

Rperry2174:
also for future viewers/testers worth noting that https://github.com/grafana/pyroscope/pull/1935 adds example in pyroscope repo works with phlare for testing these kings of things

petethepig (approved):
LGTM, found a nit and a potential bug.

docs/sources/operators-guide/configure/reference-configuration-parameters/index.md
```diff
@@ -191,6 +191,11 @@ limits:
   # CLI flag: -validation.max-label-names-per-series
   [max_label_names_per_series: <int> | default = 30]
 
+  # The tenant's shard size used by shuffle-sharding. Must be set both on
+  # ingesters and distributors. 0 disables shuffle sharding.
+  # CLI flag: -distributor.ingestion-tenant-shard-size
+  [ingestion_tenant_shard_size: <int> | default = 0]
```
eh-am, line 197: wait where did this come from?

Rperry2174: you didnt add it?

PR #624, opened 2023-04-14
Parse from until timestamps in pyroscope render call correctly.

simonswine:


simonswine:
Unsure why I created two PRs this has been merged in #625

PR #626, opened 2023-04-14
feat: comparison page

eh-am:
Screenshot 2023-04-14 at 11-34-28 Comparison process_cpu samples count cpu nanoseconds{function fast } and process_cpu samples count cpu nanoseconds{function slow } Pyroscope

Not fully useful since there's no timelines yet.
Also the only way to access it right now is to visit the page directly. Will add a sidebar after all the pages are done.


Rperry2174 (approved):
lgtm!

Issue #612, opened 2023-04-04
Cleanup blocks has unexpected behaviour when using multiple tenats


simonswine:
As we currently do not query from object store at all, we rely on the local disk to query blocks from. In order to avoid disks filling up we implemented a clean up method that becomes effective if the disk has high utilization:

https://github.com/simonswine/phlare/blob/4b3c7f639b61c51cf9c846a6c4dea5913deea758/pkg/phlaredb/phlaredb.go#LL181C6-L181C6

As this method is run once per tenant it will not handle certain cases correctly:

* An old tenant that no longer receives traffic (and hence has no active instance of PhlareDB cleaning up old blocks, will never get cleaned up.
* With multiple tenants it will delete the oldest blocks each, which could mean for a just onboarded tenant it will even delete a very recent block

#### Ideas

I am not too opinionated how we solve that but we could:

* Implement a configured retention period where we only ever delete blocks older than it.
* Move the cleanup loop outside of PhlareDB and run it once for all tenants. Only cleanup the oldest blocks across tenants until the disk is no longer in HighDiskUtilization.

PR #628, opened 2023-04-14
Add support for global size-based retention policy

kolesnikovae:
Resolves https://github.com/grafana/phlare/issues/612.

One of the drawbacks is that mount points and symlinks are not honoured: if one uses a separate file system for a specific tenant (or a group), this fact is completely ignored. An ultimate solution to this would be tracking each file system independently. As a workaround, we could make it so that a user can exclude tenants from this global policy.

This change also allows implementing per-tenant retention policies in an easy way.


pkg/ingester/retention_test.go
```diff
@@ -0,0 +1,183 @@
...
+	dir  bool
+}
+
+func (f *fakeFile) Name() string               { return f.name }
+func (f *fakeFile) IsDir() bool                { return f.dir }
+func (f *fakeFile) Info() (fs.FileInfo, error) { panic("not implemented") }
+func (f *fakeFile) Type() fs.FileMode          { panic("not implemented") }
+
+/*
+func TestPhlareDB_cleanupBlocksWhenHighDiskUtilization(t *testing.T) {
```
kolesnikovae, line 45: To be added

pkg/phlaredb/phlaredb.go
```diff
@@ -273,6 +152,9 @@ func (f *PhlareDB) loop() {
 			f.runBlockQuerierSync(ctx)
 		case <-blockScanTicker.C:
 			f.runBlockQuerierSync(ctx)
+		case b := <-f.evictCh:
+			b.evicted, b.err = f.blockQuerier.evict(b.blockID)
+			close(b.done)
```
kolesnikovae, line 157: Not related to the PR, but: to me it's not entirely clear whether the head block is visible to queries in-between `Flush` and `runBlockQuerierSync`:
```go
		case <-f.Head().flushCh:
			if err := f.Flush(ctx); err != nil {
				level.Error(f.logger).Log("msg", "flushing head block failed", "err", err)
				continue
			}
			f.runBlockQuerierSync(ctx)
```

pkg/ingester/ingester.go
```diff
@@ -97,36 +99,42 @@ func New(phlarectx context.Context, cfg Config, dbConfig phlaredb.Config, storag
...
-	}
-
-	err = i.lifecycler.AwaitRunning(ctx)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return services.StartManagerAndAwaitHealthy(ctx, i.subservices)
 }
```
kolesnikovae, line 115: I failed to find why we need to use the background context, and the server seems to be working fine without it. But I'm not 100% sure if there's a catch

pkg/phlaredb/phlaredb.go
```diff
@@ -459,3 +342,23 @@ func (f *PhlareDB) Flush(ctx context.Context) error {
...
+// No evictions should be done after and during the Close call.
+func (f *PhlareDB) Evict(blockID ulid.ULID) (bool, error) {
+	e := &blockEviction{
+		blockID: blockID,
+		done:    make(chan struct{}),
+	}
+	f.evictCh <- e
+	<-e.done
+	return e.evicted, e.err
+}
```
kolesnikovae, line 364: The idea here is to prevent disk access races: when a block is getting removed from the disk, we need to make sure that no queries will hit it. Given that we also need to manage a global retention policy the method is public

pkg/ingester/retention.go
```diff
@@ -0,0 +1,225 @@
...
+
+	"github.com/grafana/phlare/pkg/phlaredb"
+	"github.com/grafana/phlare/pkg/phlaredb/block"
+	diskutil "github.com/grafana/phlare/pkg/util/disk"
+)
+
+const (
+	defaultMinFreeDisk                        = 10 * 1024 * 1024 * 1024 // 10Gi
+	defaultMinDiskAvailablePercentage         = 0.05
+	defaultRetentionPolicyEnforcementInterval = 5 * time.Minute
```
kolesnikovae, line 26: I think it might be worthy to make the parameters configurable

simonswine: I tend to agree, it will always be a balance between flexibility and making Pyroscope complex to operate, but being able to change that or disable deletion all together is more than a nice to have. That said I wouldn't mind if you do this in here or in a follow up PR.

pkg/ingester/retention.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return err
+	}
+
+	for volumeStatsCurrent.HighDiskUtilization && len(blocks) > 0 && ctx.Err() == nil {
+		// When disk utilization is not lower since the last loop, we end the
+		// cleanup there to avoid deleting all blocks when disk usage reporting
+		// is delayed.
+		if volumeStatsPrev != nil && volumeStatsPrev.BytesAvailable >= volumeStatsCurrent.BytesAvailable {
+			level.Warn(e.logger).Log("msg", "disk utilization is not lowered by deletion of a block, pausing until next cycle")
```
simonswine, line 195:
```suggestion
	defaultRetentionPolicyEnforcementInterval = 5 * time.Minute
```

Just a note here, for my own Laptop, the filesystem `btrfs` is fairly slow to reflect the free disk change (or doesn't at all because of snapshots). So that basically means we would delete a block every 5 minutes. Which I think is ok and better than deleting all blocks because the free disk space doesn't go down.

simonswine (approved):
LGTM. Great work! I have tested this extensively locally and it did behave exactly as I expected it to. Also good effort on covering it all with tests.

pkg/ingester/retention.go
kolesnikovae: Oh, that's a great point, which worth mentioning in docs. As a small optimization, I'll implement batch removal that relies on the actual block size on disk.

kolesnikovae:
Thank you so much for the review and help with testing @simonswine!

PR #629, opened 2023-04-14
chore: replace 'Application' for 'ProfileID' to better match phlare's semantics

eh-am:
image

Honestly not sure about using "Proflie ID", any thoughts @Rperry2174?

Edit: used "Profile Type" as suggested


simonswine (approved):
LGTM


I probably would call it "Profile Type"

Rperry2174:
> LGTM
> 
> 
> 
> 
> 
> I probably would call it "Profile Type"
> 
> 

Agreed

PR #634, opened 2023-04-18
feat: truncation when converting to flamebearer

petethepig:
Tested this on OG pyroscope cloud data, the first two are with truncation, second two without truncation:

Screenshot 2023-04-17 at 9 37 03 PM

This is still work in progress, I want to run more tests, particularly curious how much faster this gets with larger responses.


pkg/querier/flamegraph.go
```diff
@@ -138,3 +144,24 @@ func ExportToFlamebearer(fg *querierv1.FlameGraph, profileType *typesv1.ProfileT
 		},
 	}
 }
+
+const maxNodesDefault = 2048
```
petethepig, line 148: I think this is the key part someone should review in this PR. I'm setting the number of nodes to 2048 by default, which can be low in some instances. There's also currently no way to override it on the frontend.

pkg/querier/flamegraph.go
```diff
@@ -45,37 +48,40 @@ func NewFlameGraph(t *tree) *querierv1.FlameGraph {
 		var i int
 		var ok bool
 		name := current.node.name
-		if i, ok = nameLocationCache[name]; !ok {
-			i = len(names)
-			if i == 0 {
-				name = "total"
+		if current.node.total >= minVal || name == "other" {
```
simonswine, line 51: Assuming the "other" is not implemented right now (maybe I missed it).

In my opinion we should be transparent, there is "more". But that could also be a follow-up PR.

api/querier/v1/querier.proto
```diff
@@ -48,6 +48,7 @@ message SelectMergeStacktracesRequest {
   string label_selector = 2;
   int64 start = 3; // milliseconds since epoch
   int64 end = 4; // milliseconds since epoch
+  int64 max_nodes = 5;
```
simonswine, line 51: Not a proto expert, but with `optional` we could get a pointer to see if it has been set or not.

```suggestion
  optional int64 max_nodes = 5;
```

simonswine, line 51: A comment would be nice here:

```suggestion
  int64 max_nodes = 5; // Limit the nodes returned to only show the node with the max_node's biggest total
```

simonswine (approved):
I am not entirely convinced this is the best way of doing that. 

It is obviously not the same for those high amount of nodes, but when I set  it to something quite low like 8. I am not getting a very helpful picture:

Screenshot from 2023-04-19 11-20-16

I think at least we should group the removed values into a "other" node on each level.

I wonder if we get better results implementing something like:

* Find the top max_nodes by self value.
* Keep all non-leaf nodes and  group the (leaf) values removed into an other node

I feel this will get quite complex quickly.

I think I would be OK to merge if we make it opt in, so default `max_nodes=-1` and expose it as an option so we can gather more feedback.

petethepig:
Ran some somewhat scientific tests.

This is all for 1 hour `{}` queries on data remote written from pyroscope og cloud self profiling to dev-001 cluster

Screenshot 2023-04-20 at 10 14 51 PM

PR #635, opened 2023-04-18
feat: timelines

eh-am:
Support timelines:
image

Couple points:
* The resolution calculation is heavily inspired by Grafana's (https://github.com/grafana/grafana/blob/bc11a484ed97d9c87e8dd42347c2c34713e9e441/pkg/tsdb/intervalv2/intervalv2.go#L1), which is more fine grained compared to OG Pyroscope's (https://github.com/grafana/pyroscope/blob/d7ed25ebb6c50a7eab906666933d48e71573d3e8/pkg/storage/segment/constants.go#L7-L24)
* Grafana's implementation uses the frontend size as a hint on how many points to calculate. 
Since OG Pyroscope doesn't do that (and changing it would require some effort), I am hardcoding a default, which can be revised later.


pkg/querier/http.go
```diff
@@ -87,8 +87,37 @@ func (q *QueryHandlers) Render(w http.ResponseWriter, req *http.Request) {
...
+			ProfileTypeID: selectParams.ProfileTypeID,
+			LabelSelector: selectParams.LabelSelector,
+			Start:         selectParams.Start,
+			End:           selectParams.End,
+			Step:          float64(timelineStep),
+		}))
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
cyriltovena, line 104: We should fire those in parallel maybe ?

cyriltovena: using `g, ctx := errgroup.WithContext(ctx)` and then Go() for each upstream call.

pkg/querier/http.go
```diff
@@ -82,13 +84,50 @@ func (q *QueryHandlers) Render(w http.ResponseWriter, req *http.Request) {
...
+	})
+
+	err = g.Wait()
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	if len(resSeries.Msg.Series) > 1 {
+		http.Error(w, fmt.Sprintf("can't construct timeline since multiple series have been returned: '%d'", len(resSeries.Msg.Series)), http.StatusBadRequest)
```
cyriltovena, line 117: Should never be the case since you're not passing a by parameters btw.

cyriltovena (approved):
LGTM

pkg/querier/http.go
eh-am: good point

PR #639, opened 2023-04-19
fix(pyroscope): use nanos sample type units for cpu & wall profiles

korniltsev:
addresses https://github.com/grafana/pyroscope/issues/2070 and https://github.com/grafana/pyroscope/issues/2073
depends on https://github.com/grafana/pyroscope/pull/1927

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -278,5 +278,6 @@ replace (
 	// merged upstream yet.
 	github.com/hashicorp/memberlist => github.com/grafana/memberlist v0.3.1-0.20220708130638-bd88e10a3d91
 	github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.41.0
+	github.com/pyroscope-io/pyroscope => ../../pyroscope/pyroscope
 	google.golang.org/grpc => google.golang.org/grpc v1.44.0
 )
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,7 @@ require (
 	github.com/prometheus/client_golang v1.14.0
 	github.com/prometheus/common v0.39.0
 	github.com/prometheus/prometheus v1.99.0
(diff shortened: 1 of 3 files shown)


dalehamel:
I was thinking of making just such a PR, but you've beaten me to it :)

I hope this can be merged and made available in a weekly release soon so we can kick the tires on it 🤞

go.mod
```diff
@@ -278,5 +278,6 @@ replace (
 	// merged upstream yet.
 	github.com/hashicorp/memberlist => github.com/grafana/memberlist v0.3.1-0.20220708130638-bd88e10a3d91
 	github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.41.0
+	github.com/pyroscope-io/pyroscope => ../../pyroscope/pyroscope
```
petethepig, line 281: make sure to fix this before merging

petethepig (approved):
LGTM, but make sure to change things in go.mod

Also, I merged https://github.com/grafana/pyroscope/pull/1927

pkg/ingester/pyroscope/ingest_adapter.go
```diff
@@ -141,7 +151,7 @@ func convertMetadata(pi *storage.PutInput) (metricName, stType, stUnit, app stri
 		stType = stTypeSamples
 		stUnit = stUnitCount
 	case "wall":
```
petethepig, line 153: This is kind of separate, but @korniltsev what do you think about defaulting to CPU if there's no prefix instead of throwing an error? (`default` case)

korniltsev: I don't have a strong opinion on this matter, I'm not opposed to this solution, I just thought that personally, I would notice an error in the logs faster than incorrect data in the UI.

PR #645, opened 2023-04-21
fix: serve UI in prod mode

eh-am:
Fixes https://github.com/grafana/phlare/issues/642

Keep in mind https://github.com/grafana/phlare/issues/638 is still outstanding, but will definitely build upon the work here.

# How it works
## Dev mode (`yarn dev`)
Backend doesn't do anything.
Basepath is `/` since it's served under `http://localhost:4040`

## Prod mode (`make build`)
Backend serves `/ui/assets` requests (css/js) using the embed file server.
All other routes are served the `index.html`, which then are handled via the frontend router (`react-router`).
When building for production, a `BASEPATH` env var is dynamically set in webpack to `/ui`. This is so that React Router knows that it needs to take that into account. This part will likely change when https://github.com/grafana/phlare/issues/638 is addressed.

PR #649, opened 2023-04-25
Make both `/ui/?query=` and `/ui?query=` work

hi-rustin:
Follow up https://github.com/grafana/phlare/pull/645

Test locally:
- Before change
image

- After change
image

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -146,7 +146,10 @@ func (a *API) RegisterAPI(statusService statusv1.StatusServiceServer) error {
 	// All assets are served as static files
 	a.RegisterRoutesWithPrefix("/ui/assets/", http.StripPrefix("/ui/", http.FileServer(uiAssets)), false, true, "GET")
 	// Serve index to all other pages
-	a.RegisterRoutesWithPrefix("/ui", uiIndexHandler, false, true, "GET")
+	a.RegisterRoutesWithPrefix("/ui/", uiIndexHandler, false, true, "GET")
+	// Redirect `/ui` to `/ui/`.
+	// See more: https://github.com/grafana/phlare/pull/649#issuecomment-1522958157.
+	a.RegisterRoute("/ui", http.RedirectHandler("/ui/", http.StatusMovedPermanently), false, true, "GET")
 
 	// register status service providing config and buildinfo at grpc gateway
 	if err := statusv1.RegisterStatusServiceHandlerServer(context.Background(), a.grpcGatewayMux, statusService); err != nil {
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -149,7 +149,7 @@ func (a *API) RegisterAPI(statusService statusv1.StatusServiceServer) error {
 	a.RegisterRoutesWithPrefix("/ui/", uiIndexHandler, false, true, "GET")
 	// Redirect `/ui` to `/ui/`.
 	// See more: https://github.com/grafana/phlare/pull/649#issuecomment-1522958157.
-	a.RegisterRoute("/ui", http.RedirectHandler("/ui/", http.StatusMovedPermanently), false, true, "GET")
+	a.RegisterRoute("/ui", http.RedirectHandler("/ui/", http.StatusFound), false, true, "GET")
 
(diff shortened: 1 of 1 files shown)


simonswine (commented):
I would prefer to redirect `/ui` to `/ui/`

hi-rustin:
> I would prefer to redirect `/ui` to `/ui/`

Usually, we use `/ui?a=xxx` instead of `/ui/?a=xxx`. For example: `https://github.com/issues?q=is%3Aopen+is%3Aissue+author%3Ahi-rustin+archived%3Afalse+phlare`.

I think adding a redirection here is a little bit weird. It is not common to have this kind of redirection.

hi-rustin:
> Honestly I didn't do any work here hoping that we would default the UI to the root instead of `/ui/` once most pages were ready. But oh well.

Off-topic: I would like to ask what is the relationship between this UI and the UI in grafana? Will they be merged? And what is the relationship with pyroscope UI?

Rperry2174:
> Off-topic: I would like to ask what is the relationship between this UI and the UI in grafana? Will they be merged? And what is the relationship with pyroscope UI?

We will provide more official guidance in this at our next community meeting / via other channels. 

However, just so you are in the loop as you've been helping us with this (thanks btw!) our long term goal is merging the Pyroscope and Phlare projects. 

Once they are merged there will likely be two ways to view profiling data:
1. Via standalone UI (the one being talked about here)
2. Via Grafana UI (explore tab and deeper integrations into Grafana)

In the meantime, we've made this UI compatible with both phlare and pyroscope projects to unify the frontend which users interact with while we work on merging the backend. 

That way when the projects are eventually merged it will be done in the "background" with minimum disruption to users (from a UI perspective)

hi-rustin:
> Once they are merged there will likely be two ways to view profiling data:
> 
> 1. Via standalone UI (the one being talked about here)
> 2. Via Grafana UI (explore tab and deeper integrations into Grafana)

Sounds great! Thanks for your reply. I'm not confused at all now.

hi-rustin:
> I don't think it is weird, usually this redirection is done to support relative linking of assets and APIs.
> 
> Maybe an example to make it clearer:
> 
> * User wants Phlare to mount a via a reverse proxy in there `https://tooling.corp/phlare`
> * When you are accessing it using `/phlare/ui`, the API will be under `../api`
> * When you are accessing it using `/phlare/ui/`, the API will be under `../../api`.

Sounds reasonable. Thanks for your review!



> Honestly I didn't do any work here hoping that we would default the UI to the root instead of `/ui/` once most pages were ready. But oh well.

Make sense. I also prefer to move UI to the root and move admin pages to `/admin` or other routes. It makes accessing the data easier.

pkg/api/api.go
```diff
@@ -147,6 +147,9 @@ func (a *API) RegisterAPI(statusService statusv1.StatusServiceServer) error {
 	a.RegisterRoutesWithPrefix("/ui/assets/", http.StripPrefix("/ui/", http.FileServer(uiAssets)), false, true, "GET")
 	// Serve index to all other pages
 	a.RegisterRoutesWithPrefix("/ui/", uiIndexHandler, false, true, "GET")
+	// Redirect `/ui` to `/ui/`.
+	// See more: https://github.com/grafana/phlare/pull/649#issuecomment-1522958157.
+	a.RegisterRoute("/ui", http.RedirectHandler("/ui/", http.StatusMovedPermanently), false, true, "GET")
```
simonswine, line 152: Nit: I think a 302 redirect has less risk, should we choose to do something else in the future

 ```suggestion
	a.RegisterRoute("/ui", http.RedirectHandler("/ui/", http.StatusFound), false, true, "GET")
```

Comment (simonswine) [inline pkg/api/api.go:152]: Nit: I think a 302 redirect has less risk, should we choose to do something else in the future

```suggestion
	a.RegisterRoute("/ui", http.RedirectHandler("/ui/", http.StatusFound), false, true, "GET")
```

simonswine (approved):
LGTM

pkg/api/api.go
hi-rustin: Thanks! Make sense.

PR #651, opened 2023-04-25
Remove `cortex_` prefix from scheduler/frontend metrics

simonswine:
This is misleading and really hard to fix later, this will also require to modify some of the existing dashboards.

In my opinion, we should also rename all metrics to use the `pyroscope_` prefix


cyriltovena (approved):
LGTM

PR #652, opened 2023-04-25
Update all metrics to be prefixed with pyroscope rather than phlare

simonswine:
This is a breaking change for all dashboards and rules.

In some future we plan to rename the project from Phlare to Pyroscope, so we rather take this hit early.


Rperry2174 (approved):
lgtm

PR #654, opened 2023-04-26
Support selectors for LabelValues/LabelNames calls

simonswine:
This adds support selectors on the `LabelNames/Values` call.

This could replace the `Series` call in Explore and improve the experience with high cardinality tenants.

Also does a fair bit of refactoring to ensure we only define one Request/Response type between ingester/querier


cyriltovena (approved):
LGTM Awesome !

PR #657, opened 2023-04-26
Removes gzip compression between querier and ingester

cyriltovena:
This is causing more CPU usage and not really useful on the read path as the data is already compressed with dictionnaries.


petethepig (approved):
:+1:

PR #621, opened 2023-04-12
Use binary search in `getSeriesIndex`

hi-rustin:
Remove TODO and use the binary search here.

PR #658, opened 2023-04-26
Use binary search in `getSeriesIndex` 2.0

simonswine:
This brings back the original pr #621, with the fix for the panic.

It makes the code quite a bit more complicated at best with questionable performance gains.

I have also submitted #659 which adds test cases to cover the actual regression.

@hi-rustin how are you feeling about not merging this


hi-rustin:
> @hi-rustin how are you feeling about not merging this

Sounds good to me. It does make the code harder to understand.

PR #659, opened 2023-04-26
Add test cases for getSeriesRef

simonswine:
This are additional test cases from the PR #621, which we reverted as a hotfix.

### DIFF
--- a/pkg/phlaredb/profile_store_test.go
+++ b/pkg/phlaredb/profile_store_test.go
@@ -2,7 +2,9 @@ package phlaredb
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


cyriltovena (approved):
LGTM

hi-rustin (approved):
Looks good to me!

I am sorry for Introducing that bug.

We may need to remove that binary search todo.

simonswine:
> Looks good to me!
> 
> I am sorry for Introducing that bug.
> 
> We may need to remove that binary search todo.

No problem at all, I also should have really caught that in the review/tests :laughing:

PR #660, opened 2023-04-26
docs: Add new client docs

Rperry2174:
Currently between Phlare and Pyroscope there exists duplicates of agents and for simplicity for users we feel like it would be best to simplify that to have **one** recommended client for each language. 

This PR eliminates the less sophisticated of the two language integrations for where two exist and brings over a new language integration for those that didn't have one in Phlare docs.


docs/sources/operators-guide/configure-agent/language-support/go.md
Rperry2174, line 1: I removed this because it felt like it didn't really align with the new structure of docs. I think anything that is in this document that we would like to keep should be moved to the `docs/sources/operators-guide/configure-agent/language-support/go_pull.md` docs.

docs/sources/operators-guide/configure-agent/language-support/go_pull.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Scrape configuration
+
+Pyroscope uses exactly the same mechanisms as Prometheus does in order to ensure smooth user experience. Therefore, it
+is configured in almost identical way, and [Prometheus documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)
+can be used as a reference.
+
+If using Pyroscope you will need to add the following content to your `pyroscope/server.yml` Pyroscope config file. See the [Server config documentation](/docs/server-configuration#configuration-file) for more information on where this config is located by default on your system.
+
+If using Phlare ... TODO
```
Rperry2174, line 33: I would be fine removing the "If using Pyroscope" section above in favor of how to use pull mode and send data to phlare -- @simonswine or @cyriltovena can you help here?

docs/sources/operators-guide/configure-agent/language-support/go_push.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+Note: If you'd prefer to use Pull mode you can find those docs here.
+
+Then add the following code to your application:
+
+```go
+package main
+
+import "github.com/pyroscope-io/client/pyroscope"
```
Rperry2174, line 28: @petethepig is this still fine after moving "github.com/pyroscope-io/client/pyroscope" to grafana org?

Rperry2174:
Also not sure how to get syntax highlighting here:
image

simonswine:
> Also not sure how to get syntax highlighting here:

@Rperry2174 I think the variant of `prism.js` that we use in grafana/website has no support for Java, maybe using `javascript` is good enough, I tried to add it myself but I struggled with the different formatting I get. 

We probably need to request `java` in `grafana/website`

Rperry2174:
> > Also not sure how to get syntax highlighting here:
> 
> @Rperry2174 I think the variant of `prism.js` that we use in grafana/website has no support for Java, maybe using `javascript` is good enough, I tried to add it myself but I struggled with the different formatting I get.
> 
> We probably need to request `java` in `grafana/website`

I was able to track it down with Roberts help: https://github.com/grafana/website/blob/master/assets/js/app/prism/river_highlighting.js

Rperry2174:
> I think you should limit us in the docs to two variants:
> 
> * push: SDK sending directly to Phlare / Grafana Cloud
> * pull: Grafana Agent pulling from `pprof` endpoints and sending it to Phlare / Grafana Cloud

For now, the only sdk which supports both is Golang. I thought about this more after we talked about it and we have a considerable amount of trafic for golang-pull relative to golang-push:
image

I think if we were to only provide those users the option to use the grafana agent for pull mode then we would no longer to be able to support many people who use golang-pull (from experience, I agree, tends to be larger companies). 

I don't think it actually ends up being confusing for users because their workflow is usually:
"I want to profile" -> language -> push or pull -> "add it to my application" 

Now with grafana added to the mix that decision will likely be made _first_ before looking at language at all. So it doesn't add a bunch of permutations like this:

"I want to profile" -> language -> push or pull-pprof **or pull-grafana-agent** -> "add it to my application" 

but instead it's more like this
image

I think this way will flow with the way user think about adding profiling to their application without leaving out the purple line people from above who use the golang pull-mode but _not_ grafana agent. 

Also this will make sure as we add pull mode for other languages we will have a place to add them that is not the grafana agent. 

Basically tldr not everyone who uses pull mode will want to use the grafana agent so I think we need to keep options for those docs long term I think we should go to a structure more like this when pull and push mode (and eventually tracing integration) are all options

image

wdyt @cyriltovena @petethepig @kolesnikovae

Rperry2174:
Removed all mention of godeltaprof not sure if there's any case where someone will need to use that directly once pull mode is in grafana agent @petethepig

PR #661, opened 2023-04-27
20230427 fix kind cluster single binary

simonswine:
- fix-single-binary-deploy
- Make dashboards work for single-binary phlare


Rperry2174 (approved):
lgtm

Issue #638, opened 2023-04-19
[ui][pyroscope] Running Pyroscope  UI behind a proxy


cyriltovena:
If we run the application behind a proxy with a path like : `https://admin-dev-us-central-0.grafana.net/directory-phlare/fire-dev-001/query-frontend/ui/?query=`

Then the queries goes to:

```
https://admin-dev-us-central-0.grafana.net/pyroscope/label-values?label=__name__
```
but should go to:

```
https://admin-dev-us-central-0.grafana.net/directory-phlare/fire-dev-001/query-frontend/pyroscope/label-values?label=__name__
```

eh-am:
Requests use
https://github.com/grafana/pyroscope/blob/main/webapp/javascript/services/base.ts#L70

Which currently depend on a custom meta tag
https://github.com/grafana/pyroscope/blob/main/webapp/javascript/util/baseurl.ts#L4-L6

We can either override the entire `@webapp/util/baseurl` file, or make the backend preprocess the `index.html` file and add the tag to the`<head>`.

simonswine:
I might just be missing the point, but why can't we set `BASE_URL` relatively like `../` ? Ideally I want this to work without a CLI flag, so that the reverse proxy can mount it under `/tools/pyrscope/ui` or under `/tooling/profiles/ui` while still using the same instance that just thinkgs it is accessed at `/ui`

eh-am:
If we set BASE_URL to `../`, if we have a SPA route like `ui/path/subpath`, then assets would be loaded from `ui/path/`, which is wrong.

---
If we are talking about being smart, we could technically (in js) get the current url and treat everything before `/ui` as the subpath.
Example:
```
http://localhost:9999/tools/pyroscope/ui 
```
Subpath is `/tools/pyroscope/`.

However this JS needs to be the first thing in the `head` (before any other assets), then run synchronously, then set the `<base>` tag dynamically.

But that's makes things slower, and it relies on always serving under a well-known path (`/ui`), which may not be what we want.

PR #663, opened 2023-04-28
feat: run under a subpath

eh-am:
Adds a `-api.base-url` flag to support running the UI under a different baseUrl.
Closes https://github.com/grafana/phlare/issues/638

# How it works
When doing a prod build (`make build`), it generates an `index.html` with `<base href={{ .BaseURL }} />` that's parsed as a go template by the phlare server and injects the flag value (with some adaptations).

It accepts absolute URLs (such as `http://localhost:8080/foobar/`) and paths `(/foobar/`).

There's some dirtiness like stripping `/ui/` when doing requests to the backend, which hopefully would be cleaned up once we move the ui to the root.

Also adds e2e tests using cypress to validate this works.

- [X] Add e2e tests
- [ ] Normalize terminology between "baseURL", "basePath", "baseName" etc

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -63,7 +63,7 @@ jobs:
         uses: cypress-io/github-action@v5
         with:
           wait-on: http://localhost:8080/foobar/
-          start: make run 'PARAMS=-base-url=/foobar/'
+          start: make run-base-url
           config-file: cypress/ci-base-path.ts
         env:
           ELECTRON_ENABLE_LOGGING: 1
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/upload-artifact@v2
         if: always()
         with:
-          name: cypress-screenshots
+          name: regular-path-cypress-screenshots
           path: cypress/screenshots
 
   base-path:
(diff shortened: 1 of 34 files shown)


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18.15.0
+          cache: yarn
+      - name: Install Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.19.8
```
simonswine, line 28: To match the other pipelines

```suggestion
          go-version: "1.19.9"
```

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,75 @@
...
```
simonswine, line 57: To match the other pipelines

```suggestion
          go-version: "1.19.9"
```

simonswine (approved):
LGTM.

Also great to get some e2e test coverage

.github/workflows/e2e.yaml
eh-am: Half of the work is just building the ui + binary :( Not sure what can be done there

simonswine: I think the `cache: true` might already shave off quite a bit

PR #665, opened 2023-05-02
ci: run 'make fmt check' concurrently

eh-am:
format step takes ~7 minutes and tests take ~7 minutes, running them serially is quite slow

For reference:

Before:
image

After:
image


cyriltovena (approved):
LGTM

PR #667, opened 2023-05-04
Make clients options non global

hi-rustin:
Done the TODO. I think it is safe to move it into the new function.


cyriltovena (approved):
LGTM

PR #668, opened 2023-05-05
 Remove unused import and fix typo

hi-rustin:
I think `public/app/example.test.tsx` is useless. We don't have any tests now.


public/app/pages/routes.ts
```diff
@@ -1,5 +1,5 @@
 export enum ROUTES {
-  CONTINOUS_SINGLE_VIEW = '/',
+  CONTINUOUS_SINGLE_VIEW = '/',
```
eh-am, line 2: Actually since there's no "adhoc" pages, the `CONTINUOUS` prefix is not needed, so maybe we should just call it `SINGLE_VIEW`?

public/app/app.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom/client';
 import './jquery-import';
 import { Provider } from 'react-redux';
-import store, { persistor } from './redux/store';
+import store from './redux/store';
```
eh-am, line 5: This is good, but doesn't prevent catching the same issue again. Would you mind adding a linter rule to check for unused imports?

public/app/pages/routes.ts
hi-rustin: Sounds good. Updated.

public/app/app.tsx
hi-rustin: Do you mean to use ESlint?  I am glad to add it. I will open a new pull request to introduce the ESlint as the linter.

eh-am: Added an issue here https://github.com/grafana/phlare/issues/672

PR #669, opened 2023-05-05
Remove useless err check in LabelValues API

hi-rustin:
I believe this is a useless check. So I removed it.

### DIFF
--- a/pkg/querier/http.go
+++ b/pkg/querier/http.go
@@ -31,19 +31,16 @@ type QueryHandlers struct {
 	upstream querierv1connect.QuerierServiceHandler
 }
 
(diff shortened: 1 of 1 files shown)


pkg/querier/http.go
```diff
@@ -31,19 +31,16 @@ type QueryHandlers struct {
...
 		return
 	}
-	var (
-		res []string
-		err error
-	)
+	var res []string
 
 	if label == "__name__" {
 		response, err := q.upstream.ProfileTypes(req.Context(), connect.NewRequest(&querierv1.ProfileTypesRequest{}))
```
cyriltovena, line 46: Something's wrong we're swallowing the error

PR #675, opened 2023-05-10
When ready check fails display service names

simonswine:
Currently the health check under `/ready` will only show the count of services in a particular state, this will make sure we will display the state and the services names.

After a recent deploy I saw a couple of replicas unavailable, be the ready endpoint did not really reveal which service was failing this adapts the output to more specific.

Before

```
Some services are not Running:
Running: 9
Starting: 2
```

After 

```
Some services are not Running:
Running: [agent distributor ingester memberlist-kv overrides-exporter query-scheduler ring server usage-stats]
Starting: [querier query-frontend]
```

### DIFF
--- a/pkg/phlare/phlare.go
+++ b/pkg/phlare/phlare.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"sort"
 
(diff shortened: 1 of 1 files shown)


cyriltovena (approved):
Nice one !

PR #676, opened 2023-05-10
feat: support multitenancy in the /ui

### LINKED ISSUE #664: Support multitenancy in the new embedded UI
This was brought up in [slack](https://grafana.slack.com/archives/C049PLMV8TB/p1682945180887389)

Currently the UI doesn't handle any multitenancy so all requests fail with a 401.

There's couple options as suggested in slack:
* Add a concept of 'datasources' similar to grafana, so that one can set those via the UI
* Add a proxy that sets the org header transparently to the UI.

Of course there are more options which need to be discussed.

eh-am:
Closes #664

Upon loading the webapp, hits the backend and checks whether multi tenancy is enabled (based on the error code/message). If so, a dialog shows up asking the user to set a Tenant ID:
image

Which is then stored in redux and hydrated from/to local storage.

Also add a sidebar item allowing the user to change tenants, while keeping existing state.
image
image


simonswine (approved):
I can see that work quite well.

I suggest we never use `Org(anization) ID`  and always use `Tenant ID` instead. This has caused so much confusion in Loki that we use `user/org-id/tenant/tenant-id` interchangeably.

Issue #672, opened 2023-05-09
[frontend] add linting


eh-am:
Would be nice to have linting rules for the frontend. We could reuse most of the ones from `@grafana/eslint-config`:
```
 {
  "extends": ["@grafana/eslint-config"]
}
```

Additional ones that would be useful:
* Forcing all imports to be absolute (using alias), this is so that they can be easily overridden downstream.
* Force React to have been imported (i think `@grafana/eslint-config` already has that, but worth mentioning)

hi-rustin:
Thanks for writing this out! I am working on this.

PR #677, opened 2023-05-11
Add eslint as linter and prettier as formatter for app

hi-rustin:
close https://github.com/grafana/phlare/issues/672

- Add eslint as the linter
- Add prettier as the formatted

Refer to https://github.com/grafana/grafana/blob/main/.eslintrc and https://github.com/grafana/pyroscope/blob/main/.prettierignore


hi-rustin:
@eh-am Could you please take a look at this PR? I added eslint as the linter and prettier as the formatter. I should have at least met the requirements of the issue. And I have GitHub Actions enabled. Thanks!

hi-rustin (commented):
Self check

scripts/webpack/webpack.common.js
```diff
@@ -48,6 +48,30 @@ module.exports = {
...
+        '../../public/app/components/SidebarTenant'
+      ),
+      '@webapp/components/SidebarTenant.module.css': path.resolve(
+        __dirname,
+        '../../public/app/components/SidebarTenant.module.css'
+      ),
+      '@webapp/components/TenantWall': path.resolve(
+        __dirname,
+        '../../public/app/components/TenantWall'
+      ),
```
eh-am, line 74: I think these are not needed, we can just add a global rule `@phlare/*` and import those as `@phlare/compoennts/MyComponent`.

.eslintrc.js
```diff
@@ -0,0 +1,55 @@
...
+    ],
+    'import/no-relative-packages': 'error',
+    'no-restricted-imports': [
+      'error',
+      {
+        patterns: [
+          {
+            group: ['../*', './*'],
+            message:
+              'Usage of relative parent imports is not allowed. Please use absolute(use alias) imports instead.',
```
eh-am, line 31: 👍

eh-am (commented):
Nice! Just one small thing, imports to this repo can be aliased to `@phlare`, it makes the distinction more clear (what's from this repo, what's from pyroscope-oss etc).

scripts/webpack/webpack.common.js
hi-rustin: Updated! 

But do we need to get rid of the name `phlare` in the future?

hi-rustin: Thanks for your review!  💚 💙 💜 💛 ❤️

eh-am: >But do we need to get rid of the name phlare in the future?
we could rename to something else instead of "phlare", but the idea is to always have absolute imports, so that downstream projects (ie projects that just import this code) can override specific components. similar to what we do in this repo regarding `pyroscope-oss`

hi-rustin: Sounds good! Thanks!

PR #701, opened 2023-05-16
Add store-gateway components

cyriltovena:
This PR adds a store-gateway component similar to what exists in Mimir. https://github.com/grafana/mimir/blob/main/pkg/storegateway/gateway.go

This is still very early, so far I've decided to depends on [Mimir to benefits from the shuffle sharding and replication strategy](https://grafana.com/docs/mimir/latest/operators-guide/architecture/components/store-gateway/).

I'm not planning to provide any block persistence for now only in-memory. In the future we should use memcached for symbols and tsdb index. The store gateway open block for each tenant within 24h - (now -2h).

Since we don't have a compactor a single gateway can download data duplicated by the replication factor of ingesters. This means we need to deduplicate blocks data while streaming now. To speed this up I've implemented a `BufferedIterator` that helps merging multiple iterator in parallel. In the future we should probably only download compacted blocks.

For now we always deduplicate even in the ingester code even though no duplication happens there, the PR was big enough and I don't think it's a big concern right now.

The store-gateway also replicate data for high availability which means more duplicates are sent to queriers. We should consider sending block ULID and request querier to select the one to consider.

To keep thing simple for the query path right now, we split the queries using the `queryStoreAfter` configuration. Ultimately if we dedupe by blocks we should be able to remove that configuration and select blocks that needs to be querier from the querier directly.

Ultimately I think we should have the store-gateway only serve blockID, index and metadata cc @kolesnikovae 

Moves forward grafana/phlare#717

Block Page on store-gateway.

image

Trace example:

image

Remaining items left for later:

- [ ] https://github.com/grafana/pyroscope/issues/2049
- [ ] https://github.com/grafana/pyroscope/issues/2050
- [ ] https://github.com/grafana/pyroscope/issues/2052
- [ ] grafana/pyroscope#2045


pkg/querier/ingester_querier.go
```diff
@@ -23,61 +29,98 @@ type IngesterQueryClient interface {
...
+	for _, r := range responses {
+		r := r
+		g.Go(util.RecoverPanic(func() error {
+			return r.response.Send(req)
+		}))
+	}
+	if err := g.Wait(); err != nil {
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
+	return responses, nil
```
kolesnikovae, line 125: I must be missing something. Here and in some other places we first collect responses from all the clients (ingesters) and only then send the request – can it be done immediately for each client (ingester), i.e in the `forAllIngesters` callback

pkg/querier/ingester_querier.go
```diff
@@ -23,61 +29,98 @@ type IngesterQueryClient interface {
...
+	})
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
+	// send the first initial request to all ingesters.
+	g, _ := errgroup.WithContext(ctx)
+	for _, r := range responses {
+		r := r
+		g.Go(util.RecoverPanic(func() error {
+			return r.response.Send(req)
```
kolesnikovae, line 119: I'm afraid but concurrent use of the same request here may cause a race condition ([see similar issue](https://github.com/grafana/phlare/pull/670))

cyriltovena: I'll probably clone it no need to take the risk

pkg/phlaredb/block_querier.go
```diff
@@ -1030,31 +1089,45 @@ func (q *singleBlockQuerier) openFiles(ctx context.Context) error {
...
 	}
 	if parquetFile.NumRows() == 0 {
 		return fmt.Errorf("error parquet file '%s' contains no rows", filePath)
 	}
-
+	opts := []parquet.FileOption{
+		parquet.SkipBloomFilters(true), // we don't use bloom filters
+		parquet.FileReadMode(parquet.ReadModeAsync),
+		parquet.ReadBufferSize(1024 * 1024),
+	}
```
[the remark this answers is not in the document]
cyriltovena: I've played with 4 MB, 2MB, 1MB and 0.5MB and it seems that 1MB is the best one for now.

kolesnikovae (approved):
LGTM

Awesome work!

Issue #689, opened 2023-05-12
SelectMerge latency is high


kolesnikovae:
Most of the queries spend significant time after all the samples were fetched and deduplicated. We need to figure out and fix what causes the latency.

For example, phlare-querier **SelectMergeStacktraces**:
![image](https://github.com/grafana/phlare/assets/12090599/947dd28f-7cb6-4c8c-8e17-9f296b0e2d4b)

Solving the problem may result in a very significant decrease in the overall query latencies (up to 50%).

It’s very likely that this is caused by pulling and merging resolved stacks from the ingesters: the size of the payload may be quite big. If this is the case, we may want to find a way to reduce it, e.g. by stack truncation:
![image](https://github.com/grafana/phlare/assets/12090599/2d8f252d-fdd0-4133-8e17-4866b6c1a8a5)

We may want to optimize the representation of results in this API: notice that stacks take appx. 50MB (encoded in protobuf). We should consider building a truncated tree (w/o insignificant nodes) as close to the storage as possible, instead of passing an array of stack traces along the way. That would also decrease CPU time of the query execution, and allocations as well:
<img width="1689" alt="image" src="https://github.com/grafana/phlare/assets/12090599/04f4b21e-f8e0-44d6-ab39-84d789932fd8">
![image](https://github.com/grafana/phlare/assets/12090599/785d2009-1357-44ff-a856-77fb7f6b56aa)

kolesnikovae:
Also related: https://github.com/grafana/pyroscope/issues/2107

cyriltovena:
Based on our discussions I think this is a great idea to use an opaque format for the internal API. We should use the Pyroscope tree package https://github.com/grafana/pyroscope/blob/254c5759900d3a2da6e6dfaf4ef2767c05cb45bb/pkg/storage/tree/tree.go between query-frontend - querier - ingester and store-gateway.

kolesnikovae:
grafana/phlare#702 reduces the query duration 2-3 times in certain cases:

![Image](https://github.com/grafana/phlare/assets/12090599/c2b777fa-454a-49cc-92bf-67ed15337551)

As a bonus, resource consumption also decreased:

![Image](https://github.com/grafana/phlare/assets/12090599/c971290d-1dbb-442a-be0c-b945aef55a5d)

Closing the issue for now

PR #702, opened 2023-05-18
Use pyroscope tree representation in MergeProfilesStacktraces API

kolesnikovae:
The PR is aimed at decreasing querier-ingester latency. See https://github.com/grafana/phlare/issues/689:
> We may want to optimize the representation of results in this API: notice that stacks take appx. 50MB (encoded in protobuf). We should consider building a truncated tree (w/o insignificant nodes) as close to the storage as possible, instead of passing an array of stack traces along the way


api/ingester/v1/ingester.proto
```diff
@@ -42,18 +42,29 @@ message SelectProfilesRequest {
...
   repeated string function_names = 2;
+  // Merge result marshaled to pyroscope tree bytes.
+  bytes tree_bytes = 4;
+}
+
+enum StacktracesMergeFormat {
+  MERGE_FORMAT_UNSPECIFIED = 0;
+  MERGE_FORMAT_STACKTRACES = 1;
+  MERGE_FORMAT_TREE = 2;
 }
```
kolesnikovae, line 68: I was thinking about using `oneof` but decided to use an explicit enum (because eventually we should use a single format).

I don't feel strong about that – please let me know if you want me to change it

pkg/phlaredb/block_querier.go
```diff
@@ -561,28 +559,29 @@ func (q Queriers) MergeProfilesStacktraces(ctx context.Context, stream *connect.
 			if err != nil {
 				return err
 			}
-			lock.Lock()
-			defer lock.Unlock()
-			result = append(result, merge)
+			m.MergeStackTraces(merge.Stacktraces, merge.FunctionNames)
```
cyriltovena, line 562: Wouldn't it be better if we return the tree directly from the `q.MergeByStacktraces` ?

pkg/model/tree.go
```diff
@@ -0,0 +1,471 @@
...
+
+type Tree struct {
+	root []*node
+}
+
+func emptyTree() *Tree {
+	return &Tree{}
+}
+
+func newTree(stacks []stacktraces) *Tree {
```
cyriltovena, line 23: is this for backward compatibility ?

pkg/phlaredb/block_querier.go
```diff
@@ -561,28 +559,29 @@ func (q Queriers) MergeProfilesStacktraces(ctx context.Context, stream *connect.
...
 		return err
 	}
 
 	// sends the final result to the client.
 	sp.LogFields(otlog.String("msg", "sending the final result to the client"))
 	err = stream.Send(&ingestv1.MergeProfilesStacktracesResponse{
-		Result: phlaremodel.MergeBatchMergeStacktraces(result...),
+		Result: &ingestv1.MergeProfilesStacktracesResult{
+			Format:    ingestv1.StacktracesMergeFormat_MERGE_FORMAT_TREE,
+			TreeBytes: m.TreeBytes(r.GetMaxNodes()),
```
cyriltovena, line 583: Nice

cyriltovena (approved):
LGTM, really nice this is backward compatible.

pkg/phlaredb/block_querier.go
kolesnikovae: This is a good point. I'd experiment with this. My idea was to merge samples to the tree (in batches) directly in `MergeByStacktraces`, and pass the merger as an argument. However, I wasn't not 100% sure if this approach is more efficient, and I didn't want to delay shipping of this feature

pkg/model/tree.go
kolesnikovae: Used in tests. Refactored helpers to `_test.go`

PR #703, opened 2023-05-19
Reorganize docs structure

Rperry2174:
Reorganizing docs structure

image

I made a couple changes to the structure to make it a little more straightforward.

I don't think it helped to nest everything under "operator and user guide" as opposed to having all the potentially options available to users immediately in the sidebar. 

I think the most common flow when people come to docs is going to be:

1. "How do I get started" (overview)
2. "How do I get started getting profiles from my code" 
3. "How do I set up the server" (if necessary -- in the case of the cloud they won't need this which is why I think client configuration is more important than server) 
4. "how do I deploy it" 

And then after that I put everything in "reference" because it's things that people can refer to if they want to, but if users are trying to get started quickly then they can still do that without the reference sections


petethepig (approved):
LGTM.

We should discuss releasing phlare so that these docs show up by default.

PR #706, opened 2023-05-23
Fix panic in usagestats

simonswine:


simonswine:
Once we are happy, I would contribute it back to loki

cyriltovena (approved):

PR #708, opened 2023-05-23
fix: only pass `X-Scope-OrgID` header if multi tenant is enabled

eh-am:
I noticed if one goes from a multi_tenant instance to a single_tenant, the orgID header would still be passed in requests. The backend seems to ignore it, but may be best to simply not pass it, since it's also misleading.

Adds tests to cover that case.

Also update jest to use same module resolution as typescript.


cyriltovena (approved):
LGTM

PR #709, opened 2023-05-24
fix(frontend): make tenant menu show even when sidebar is collapsed

eh-am:
When multitenancy is enabled and sidebar is collapsed, the menu is being hidden.
Screenshot 2023-05-24 at 10 19 57

This PR solves it by rendering the component outside the normal flow (using Portal), next to the body. Reason is that the sidebar itself contains `overflow: hidden`, which if we disable provides other issues (like a weird scrollbar)
Screenshot 2023-05-24 at 11 03 01


cyriltovena (approved):
LGTM

PR #710, opened 2023-05-24
fix: rename pyroscope_app to `service_name`

korniltsev:


cyriltovena (approved):
LGTM

pkg/ingester/pyroscope/ingest_adapter.go
```diff
@@ -99,7 +99,7 @@ func (p *pyroscopeIngesterAdapter) Put(ctx context.Context, pi *storage.PutInput
 		Name:  labels.MetricName,
 		Value: metric,
 	}, &typesv1.LabelPair{
-		Name:  "pyroscope_app",
+		Name:  "__service_name__",
```
cyriltovena, line 102: I think we should do `service_name` now that I'm thinking about it because __ are not meant to be shown at query time.

cyriltovena (approved):
LGTM

PR #712, opened 2023-05-24
feat(frontend): new app selector

eh-am:
Related https://github.com/grafana/pyroscope/issues/2055

# Example (updated)
https://github.com/grafana/phlare/assets/6951209/bbf85e45-201d-4f04-b528-61236df97a05

It also supports mixing and matching `pyroscope_app` and `__service_name__`:

https://github.com/grafana/phlare/assets/6951209/c429e027-0943-45e2-87d5-82b7cacaf2c8

# Features
* It adds an Application Selector which leads user to make a more targeted query by default, similar to what OG Pyroscope does. 
* If the user still wants to write their own (complex) queries, it's still possible.
* It supports both `pyroscope_app` and `__service_name__` as "indexes"

# Caveats/Limitations
* It uses the `/querier.v1.QuerierService/Series` to create a list of apps, which returns more data than needed.
* Also it only returns data that's in memory, ie. recently ingested apps. Related to https://github.com/grafana/pyroscope/issues/2071
* Parsing an "App" from a query does not work when the query is non trivial (eg uses of `!~` etc). In this PR it doesn't make too much difference, as the only thing that happens is not populating the dropdown correctly (which to be honest, it shouldn't, after all, a query like `cpu{app="myapp.*"}` doesn't really match to an "App"
* It does not preserve tags when changing apps. For example, given a current query `cpu{mytag="foo", pyroscope_app="myapp"}` and clicking on `myapp2`, even if `myapp2` shares the exact tags, the new query will completely strip `mytag="foo"`: `cpu{pyroscope_app="myapp2"}`
* The `pyroscope_app`/`__service_name__` tag is present in the app selector.
* There's no filtering mechanism like in OG Pyroscope.


public/app/overrides/services/apps.ts
```diff
@@ -0,0 +1,87 @@
...
+  // TODO: is this the best query?
+  const response = await requestWithOrgID('/querier.v1.QuerierService/Series', {
+    method: 'POST',
+    body: JSON.stringify({
+      matchers: [],
+    }),
+    headers: {
+      'content-type': 'application/json',
+    },
+  });
```
eh-am, line 80: @simonswine I wonder if there isn't a better way to get this data...

eh-am: Not quite, more like if we shouldn't create an endpoint that does most of this work (create an "App" that combines most of its labels). Plus I don't know how bad it currently is performance wise.

simonswine: I am unsure what the method is supposed to achieve.

* Do you just want to get a flat list of unique values for `pyroscope_app` (or soon `__service_name__`): Then you should rather use LabelValues: 
  ```curl -vL -d '{"name":"__service_name__"}' -X POST -H "content-type: application/json" http://localhost:4100/querier.v1.QuerierService/LabelValues```
  It also supports matchers just like Series
* Or do you need the full label set?: Then series is the way to go

eh-am: I think I need the full label set, since I need `__service_name__` + `profileID` (to be able to query) + other metadata fields, like `__type__` and `__name__` (to be able to display more info about what you are selecting) :(

Rperry2174 (approved):
LGTM!!! Thanks @eh-am

Issue #714, opened 2023-05-25
Pyroscope ingest endpoint status code


cyriltovena:
When a user is throttled we actually don't properly propagate [status code](https://admin-dev-us-central-0.grafana-dev.net/grafana/explore?left=%7B%22datasource%22:%22P09658BD21492D183%22,%22queries%22:%5B%7B%22expr%22:%22100%20%2A%20sum%20by%20%28namespace,%20job,%20route,%20cluster%29%20%28rate%28pyroscope_request_duration_seconds_count%7Bstatus_code%3D~%5C%225..%5C%22%7D%5B1m%5D%29%29%20%2F%20sum%20by%20%28namespace,%20job,%20route,%20cluster%29%20%28rate%28pyroscope_request_duration_seconds_count%5B1m%5D%29%29%20%3E%201%22,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22cortex-dev-01-dev-us-central-0%22%7D,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22range%22:true,%22instant%22:true%7D,%7B%22refId%22:%22B%22,%22expr%22:%22100%20%2A%20sum%20by%20%28namespace,%20job,%20route,%20cluster,status_code%29%20%28rate%28pyroscope_request_duration_seconds_count%7Bstatus_code%3D~%5C%225..%5C%22%7D%5B1m%5D%29%29%20%2F%20sum%20by%20%28namespace,%20job,%20route,%20cluster,status_code%29%20%28rate%28pyroscope_request_duration_seconds_count%5B1m%5D%29%29%20%3E%201%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22cortex-dev-01-dev-us-central-0%22%7D,%22editorMode%22:%22code%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D&orgId=1).

![Image](https://github.com/grafana/pyroscope-squad/assets/1053421/ba289aa2-77d9-4263-89ea-e03b52c9bc33)

This burns our SLO and should be fixed.

I suspect this might also be the case in case for any 4xx.

cc @korniltsev if you have time to take a look.

cyriltovena:
@korniltsev we fixed this ? I still see 500 in dev:

[...]

PR #718, opened 2023-05-29
fix(pyroscope ingest): propagate connect-go http status codes

korniltsev:
address #714


pkg/ingester/pyroscope/connect.go
```diff
@@ -0,0 +1,46 @@
+package pyroscope
+
+import (
+	"github.com/bufbuild/connect-go"
+)
+
+func connectCodeToHTTP(code connect.Code) int {
```
kolesnikovae, line 7: Should we use `pkg/util/connectgrpc.CodeToHTTP` here?

kolesnikovae (approved):
LGTM

pkg/ingester/pyroscope/connect.go
korniltsev: yes!

PR #721, opened 2023-05-30
feat: select first app automatically

eh-am:
Depends on https://github.com/grafana/phlare/pull/712

When the application first starts, it loads the list of apps, and if available, sets the first one as the current query. This is so that users see something "useful" right away.

If user deletes the query later, it won't do anything (to not be annoying).

https://github.com/grafana/phlare/assets/6951209/16911955-20ce-4c95-ac72-5afc3cbb7577

# Limitations
It doesn't update history in place, so if one hits the back button, it goes back to an empty query.


Rperry2174 (approved):
LGTM!

PR #723, opened 2023-05-30
Fix stacktrace tree min value counting

kolesnikovae:
In certain cases an attempt to find the min value for a tree causes panic


cyriltovena (approved):
LGTM

Issue #722, opened 2023-05-30
Routing doesn't keep query params


eh-am:
As illustrated by the video below, routing is weird:
* upon first changing routes, it keeps the query params but doesn't change the path
* upon changing routes again, it removes query params 

https://github.com/grafana/phlare/assets/6951209/266a9761-bd82-474b-b9b8-a35fb968b057

PR #737, opened 2023-05-31
fix: routing

eh-am:
Closes https://github.com/grafana/phlare/issues/722

Funnily enough I couldn't reproduce in CI, but trust me ~bro~, the issue is indeed happening.

But basically for `redux-query-sync` to work, the same `history` instance needs to be shared across the app. Didn't dive too deep, but it was affecting how other components that use the `useLocation` were rendered. Basically `Sidebar` wouldn't rerender when the route changes, therefore its navigation items wouldn't include the query parameter!


cyriltovena (approved):
LGTM

Issue #727, opened 2023-05-31
Only start ui server when all services are ready in single binary mode


eh-am:
Otherwise one may try to access the UI (`http://localhost:4100/ui`) and get error messages saying that the frontend is not running yet:

```
level=warn caller=gokit.go:79 caller=http.go:196 traceID=17ac6d89e8d65d4b msg="GET /pyroscope/label-values?label=__name__ (500) 91.625µs Response: \"unavailable: grpc roundtripper error: frontend not running: New\\n\" ws: false; Accept: */*; Accept-Encoding: gzip, deflate, br; Accept-Language: en-US,en;q=0.5; Connection: keep-alive; Referer: http://localhost:4100/ui/?query=; Sec-Fetch-Dest: empty; Sec-Fetch-Mode: cors; Sec-Fetch-Site: same-origin; User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/113.0; X-Scope-Orgid: ; "
```

Also that makes e2e tests more difficult, since it assumes the UI is only available when everything else is also available.

eh-am:
Actually since this is such an edge case, may be best to just fix it for E2E tests.

PR #738, opened 2023-06-01
ci: [e2e] wait for /ready instead of hitting the ui directly

eh-am:
Closes https://github.com/grafana/phlare/issues/727

It waits on `/ready` instead of hitting `/ui`. There's still a chance that we try to run tests while the query-frontend is being initialized, but it's already a better heuristic than simply hitting the /ui.


hi-rustin (approved):
Looks good to me!

simonswine (approved):
LGTM

PR #745, opened 2023-06-05
chore: Ensure presence of service_name label 

korniltsev:
https://github.com/grafana/pyroscope/issues/2055


pkg/agent/profiles.go
```diff
@@ -167,7 +169,15 @@ func populateLabels(lset labels.Labels, cfg ScrapeConfig) (res, orig labels.Labe
 		lb.Set(model.InstanceLabel, addr)
 	}
 
+	if serviceName := lset.Get(phlaremodel.LabelNameServiceName); serviceName == "" {
+		if serviceName = lset.Get(phlaremodel.LabelNameServiceNameK8s); serviceName == "" {
+			return nil, nil, errors.New("service_name label is not specified")
```
petethepig, line 174: So if I don't have either of these it will throw an error on ingestion, right?

Maybe this was discussed already and there's consensus, in which case ignore me, but I would prefer we default to some values like "unspecified" or "default" or something like that.

cyriltovena (approved):
LGTM

I think we might want to also add a pyrosscope to our default config file  which scrape itself WDYT ?

PR #748, opened 2023-06-05
docs: Update links to point at grafana.com instead of pyroscope.io

Rperry2174:
Updates links to point at grafana.com instead of pyroscope.io


cyriltovena (approved):
LGTM

Issue #690, opened 2023-05-12
Stack trace symbols resolution is slow


kolesnikovae:
It’s been noticed that **SelectMergeStacktraces** queries typically spend significant (~30%) amount of time in resolving stacktraces. This is especially evident when an on-disk block is queried.

Example: in the screenshot below you can see that we've spent 2 seconds reading ~630k stacktraces from the parquet table (appx. 500 pages) – it's three times longer than fetching sample data itself:
![image](https://github.com/grafana/phlare/assets/12090599/93423f1a-8124-4244-8cb1-330fe5eaa066)

One of the hypotheses is that data locality of the stacktraces parquet table may be poor in certain cases, which results in high read amplification, therefore the column iterator needs to scan the whole file (seeking page by page). We need to investigate the problem further and optimize stacktraces parquet table, if necessary:
- Try to order rows by Series ID first – this should improve data locality: stack traces are typically fetched for a single application, or a subset of them.
- Find the optimal page size. Experiment with read ahead access.
- Encoding: figure out if Delta encoding is appropriate for the LocationIDs array and fix it if not. Potentially, this optimization only decreases the CPU time spent in varint decoding. 
- Model: use uint32 for LocationIDs and StacktraceID: 4 billions is enough to reference locations within a single block. This optimization probably only improves in-memory representation, on-disk size is unlikely to change.

We also may want to experiment with stack truncation, e.g keep only those that account for 95-99% of values, and create a stub frame for the truncated ones. This should be done before resolving locations (reading stacktraces parquet table). This is meaningful only if the relevant stacktraces are stored close to each other on disk, otherwise the amount of rows/pages we need to fetch will change insignificantly, and the latency of the operation will not decrease.

One approach is to return the result of aggregation by addressing stacktraces globally, and let the querier decide what to keep. Another way is to truncate stacks locally in ingesters, but the resulting flamegraph may be less accurate.

kolesnikovae:
Resolved via #767

PR #749, opened 2023-06-06
WIP: Shard stacktraces table by service name

kolesnikovae:
Note that this PR is experimental and is not supposed to be merged. In the PR I explore an idea of explicitly sharded `stacktraces` parquet table, where the sharding key is the `service_name`. The aim of the sharding is to improve data locality of stack traces produced by the same program (application/service) on disk: see https://github.com/grafana/phlare/issues/690 for the context.

---

Probably, the actual implementation will adhere to a more generic approach by utilising `Mapping` profile member:

```proto
  // Mapping from address ranges to the image/binary/library mapped
  // into that address range.  mapping[0] will be the main binary.
  repeated Mapping mapping = 3;
```

<details>
  <summary>Mapping</summary>
  
```proto
message Mapping {
  // Unique nonzero id for the mapping.
  uint64 id = 1;
  // Address at which the binary (or DLL) is loaded into memory.
  uint64 memory_start = 2;
  // The limit of the address range occupied by this mapping.
  uint64 memory_limit = 3;
  // Offset in the binary that corresponds to the first mapped address.
  uint64 file_offset = 4;
  // The object this entry is loaded from.  This can be a filename on
  // disk for the main binary and shared libraries, or virtual
  // abstractions like "[vdso]".
  int64 filename = 5;  // Index into string table
  // A string that uniquely identifies a particular program version
  // with high probability. E.g., for binaries generated by GNU tools,
  // it could be the contents of the .note.gnu.build-id field.
  int64 build_id = 6;  // Index into string table

  // The following fields indicate the resolution of symbolic info.
  bool has_functions = 7;
  bool has_filenames = 8;
  bool has_line_numbers = 9;
  bool has_inline_frames = 10;
}
```
  
</details>

### DIFF
--- a/pkg/phlaredb/block/block.go
+++ b/pkg/phlaredb/block/block.go
@@ -18,6 +18,8 @@ import (
 )
 
 const (
+	SymbolsShardsFilename = "symbols_shards.json"
+
 	IndexFilename        = "index.tsdb"
 	ParquetSuffix        = ".parquet"
 	DeletionMarkFilename = "deletion-mark.json"
--- a/pkg/phlaredb/block_querier.go
+++ b/pkg/phlaredb/block_querier.go
@@ -3,6 +3,7 @@ package phlaredb
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"math"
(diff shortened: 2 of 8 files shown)


kolesnikovae:
Work will continue in https://github.com/grafana/phlare/pull/767

PR #750, opened 2023-06-06
fix: /pyroscope/ingest return 400 in case of bad request

korniltsev:
This is a second attempt to fix `500` in `/pyroscope/ingest` https://github.com/grafana/phlare/issues/714

- if the request is bad and we can't convert to pprof - return `connect.CodeInvalidArgument` which will become `400`
- if the app has no profile type suffix - assume it is "cpu" and not treat as error


simonswine (approved):
LGTM

PR #752, opened 2023-06-06
Inline correctly the store-gateway config

cyriltovena:
This was a mistake from me, the config should be inline to avoid having the property `config` below `store_gateway`


kolesnikovae (approved):
LGTM

PR #757, opened 2023-06-07
Experimenting with storing stacktraces differently

simonswine:
With our given problems around retrieving esp. stacktraces in a reasonable amount of time. I have been experimenting a bit.

Two ideas:

* Rather than storing the full location ID list us a self referencing schema.
* Make use of parquet's nested object to keep track of the ids.

I will report back if any of the provide more tangible results. 

This is the current size comparision between status quo and the self referencing model:

image


cyriltovena (commented):
That's interesting makes me wonder if we should explore a custom format for storing this.

simonswine:
Superseeded by #767

PR #760, opened 2023-06-08
Flush profiles rowgroup asynchronously

cyriltovena:
This PR makes flushing row groups of profiles asynchronously and outside of the read path.

Only one flushing can happen at once. This should drastically reduce the tail latency.

image

While working on this I realised we have a race between queries getting queriers (head and disk) and flushing row groups

I believe this could be the source of:
- https://github.com/grafana/pyroscope/issues/2054
- https://github.com/grafana/pyroscope/issues/2061

I decided to not fix this in this PR as it requires a larger refactoring.

### DIFF
--- a/pkg/phlaredb/profile_store.go
+++ b/pkg/phlaredb/profile_store.go
@@ -53,16 +53,25 @@ type profileStore struct {
 	rowsFlushed uint64
 	rowGroups   []*rowGroupOnDisk
 	index       *profilesIndex
+
+	flushing    *atomic.Bool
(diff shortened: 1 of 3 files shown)


cyriltovena:
Deployed in dev for an hour or so 
image

kolesnikovae:
> While working on this I realised we have a race between queries getting queriers (head and disk) and flushing row groups

Do you have a stack trace of the race? I'd like to look into it

kolesnikovae (approved):
LGTM! 👏🏻

PR #763, opened 2023-06-09
Fix concurrent map access in ingester

cyriltovena:
Introduced by https://github.com/grafana/phlare/commit/faf62ed434feef64d4db61797b3f92659b4b55e2 this was because before we were not flushing async and we would only sort/add profiles one at a time.


kolesnikovae (approved):

korniltsev:
Wonder if we should add --race to the test runner

cyriltovena:
So this has re-introduced a bit of latency not as much as before though. I'll investigate if there's a better way before mergning

cyriltovena:
Latest commit bring back tail latency to 50ms see below

image

PR #775, opened 2023-06-16
Introduce stacktrace partition

simonswine:
This determines the partition of a particular profile, by looking first
at its metadata:

* If there is a `Filename` on the main mapping use its
  filepath.Base(Filename)
* Failing that take the externally supplied `service_name`
* Fallback to `unknown`

Take the underlying string value and hash.


cyriltovena (approved):
LGTM

PR #778, opened 2023-06-19
Increase goreleaser timeout

kolesnikovae:
On average, the release process takes a little over half an hour (see https://github.com/grafana/phlare/actions/workflows/weekly-release.yml), and the default goreleaser timeout is 30m.

I think it makes sense to increase goreleaser timeout to 60m. Long term we should try to decrease the job duration.


simonswine (approved):
LGTM

PR #779, opened 2023-06-19
Reduce join_members for memberlist using dns+

cyriltovena:
This will reduce the number of members to join at startup. Before it would try to join all members, now using `dns+` it will only select a subset.

See https://github.com/grafana/loki/pull/9723


simonswine (approved):
Good catch! LGTM

PR #780, opened 2023-06-19
Lazy block head initialization

kolesnikovae:
See https://github.com/grafana/pyroscope/issues/2047 for context. This change makes PhlareDB to initialise the head lazily at the first ingestion request.

### DIFF
--- a/pkg/phlaredb/head.go
+++ b/pkg/phlaredb/head.go
@@ -17,7 +17,6 @@ import (
 	"github.com/gogo/status"
 	"github.com/google/pprof/profile"
 	"github.com/google/uuid"
-	"github.com/grafana/dskit/multierror"
 	"github.com/opentracing/opentracing-go"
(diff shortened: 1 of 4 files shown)


pkg/phlaredb/head.go
```diff
@@ -847,23 +846,12 @@ func (h *Head) Series(ctx context.Context, req *connect.Request[ingestv1.SeriesR
...
-
-	var merr multierror.MultiError
-	for _, t := range h.tables {
-		merr.Add(t.Close())
-	}
-
-	h.wg.Wait()
-	return merr.Err()
-}
-
```
kolesnikovae, line 862: I removed the method because it is impossible to use it correctly:
1. `Close` is supposed to be called after `Flush`, and tables are closed at `Flush`, therefore the call always return an error. It worked, because we always created an empty head right away after flush.
2. There is no case when head is just closed without flushing data.

cyriltovena (approved):
LGTM

PR #785, opened 2023-06-21
Add block opened metrics

cyriltovena:
This add the number of block currently opened and also the total samples in memory.

### DIFF
--- a/pkg/phlaredb/metrics.go
+++ b/pkg/phlaredb/metrics.go
@@ -29,6 +29,7 @@ type headMetrics struct {
 
 	sampleValuesIngested *prometheus.CounterVec
 	sampleValuesReceived *prometheus.CounterVec
+	samples              prometheus.Gauge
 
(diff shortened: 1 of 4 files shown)


pkg/phlaredb/metrics.go
```diff
@@ -201,8 +207,13 @@ func newBlocksMetrics(reg prometheus.Registerer) *blocksMetrics {
 			Name: "pyroscopedb_block_opening_duration",
 			Help: "Latency of opening a block in seconds",
 		}),
+		blockOpened: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "pyroscopedb_block_opened",
```
simonswine, line 211: The name is not very clear

```suggestion
			Name: "pyroscopedb_blocks_open",
```

or 

```suggestion
			Name: "pyroscopedb_blocks_currently_open",
```

simonswine (approved):
LGTM, apart from that one metric name

PR #786, opened 2023-06-21
Removes pprof labels since we don't use them currently.

cyriltovena:
Pretty sure we will re-introduce this later, but right now it's consuming a lot of machine words in memory for nothing. I suspect at least ~1GB per pods.


cyriltovena:
This is not backward compatible so causing issue when deploying.

PR #798, opened 2023-06-26
Increase Count per RowGroup by 2

cyriltovena:
Since we now can ingest more data in memory, we should try to avoid cutting too many row groups since they are not sorted across them.

Testing this in dev.

image


cyriltovena:
Will revisit later this hasn't really made the impact I wanted.

PR #801, opened 2023-06-27
Improve Stacktraces Samples Memory Layout

cyriltovena:
This introduce a new way to abstract away the memory representation and the file format on disk (parquet). The implementation heavily relies on `parquet.RowReader`

```
// RowReader reads a sequence of parquet rows.
type RowReader interface {
	// ReadRows reads rows from the reader, returning the number of rows read
	// into the buffer, and any error that occurred. Note that the rows read
	// into the buffer are not safe for reuse after a subsequent call to
	// ReadRows. Callers that want to reuse rows must copy the rows using Clone.
	//
	// When all rows have been read, the reader returns io.EOF to indicate the
	// end of the sequence. It is valid for the reader to return both a non-zero
	// number of rows and a non-nil error (including io.EOF).
	//
	// The buffer of rows passed as argument will be used to store values of
	// each row read from the reader. If the rows are not nil, the backing array
	// of the slices will be used as an optimization to avoid re-allocating new
	// arrays.
	//
	// The application is expected to handle the case where ReadRows returns
	// less rows than requested and no error, by looking at the first returned
	// value from ReadRows, which is the number of rows that were read.
	ReadRows([]Row) (int, error)
}

```

it uses this new abstraction to represent stacktraces samples differently.

Instead of using a slice of struct such as :

```
type Profile {
   Samples []*Samples
}
type Sample struct {
	StacktraceID uint64             `parquet:",delta"`
	Value        int64              `parquet:",delta"`
	Labels       []*profilev1.Label `parquet:",list"`
}
```

It uses a double slice :

```
type Profile {
   Samples Sample
}
type Samples struct {
	StacktraceIDs []uint32
	Values        []uint64
}
```

This highly reduce the amount of memory while ingesting profiles since we use less adress space. On top of that we don't use reflection anymore when flushing Profiles by using a custom parquet serialisation.

This is running in dev and has reduce memory usage by 50% 
image


pkg/phlaredb/schemas/v1/profiles.go
```diff
@@ -124,3 +127,242 @@ func (*ProfilePersister) Reconstruct(row parquet.Row) (id uint64, s *Profile, er
...
+}
+
+func NewInMemoryProfilesRowReader(slice []InMemoryProfile) *SliceRowReader[InMemoryProfile] {
+	return &SliceRowReader[InMemoryProfile]{
+		slice:     slice,
+		serialize: DeconstructMemoryProfile,
+	}
+}
+
+func DeconstructMemoryProfile(imp InMemoryProfile, row parquet.Row) parquet.Row {
```
simonswine, line 299: Great work! I do think there might be some minor difference in how the optional fields are treated, but I don't think it will make a major difference. And it can be quite tricky to get the definition levels right:

```
diff --git a/pkg/phlaredb/schemas/v1/profiles_test.go b/pkg/phlaredb/schemas/v1/profiles_test.go
index 02ff9e7d..3bb53be6 100644
--- a/pkg/phlaredb/schemas/v1/profiles_test.go
+++ b/pkg/phlaredb/schemas/v1/profiles_test.go
@@ -41,6 +41,22 @@ func TestRoundtripProfile(t *testing.T) {
        require.NoError(t, err)
        require.Equal(t, expected, actual)

+       t.Run("EmptyOptionalField", func(t *testing.T) {
+               profiles := generateProfiles(1)
+               for _, p := range profiles {
+                       p.DurationNanos = 0
+               }
+               inMemoryProfiles := generateMemoryProfiles(1)
+               for i := range inMemoryProfiles {
+                       inMemoryProfiles[i].DurationNanos = 0
+               }
+               expected, err := readAll(NewProfilesRowReader(profiles))
+               require.NoError(t, err)
+               actual, err := readAll(NewInMemoryProfilesRowReader(inMemoryProfiles))
+               require.NoError(t, err)
+               require.Equal(t, expected, actual)
+       })
+
        t.Run("EmptyComment", func(t *testing.T) {
                profiles := generateProfiles(1)
                for _, p := range profiles {

```

simonswine, line 299: Maybe worth keep scope to the package:

```suggestion
func deconstructMemoryProfile(imp InMemoryProfile, row parquet.Row) parquet.Row {
```

simonswine (approved):
LGTM. Some great improvements

PR #767, opened 2023-06-14
Introduce symdb

kolesnikovae:
For context, please refer to [the proposal](https://docs.google.com/document/d/1ypIgFHc5szo91aWmBZebkg9RgvjyjCaJPU5JO28jyiU/edit).

TODO:
 - [x] (@simonswine) Add **`mapping_name`** field to the [Profile schema](https://github.com/grafana/phlare/blob/48fa3925eb840819a5e63c695dc67697be792451/pkg/phlaredb/schemas/v1/profiles.go#L45) and ensure it is present at read time (merge step):
    * The field identifies the main binary of the profiled program and all its versions. It is supposed to be used as the partitioning key of the stack trace table:
      * Use `Mapping.filename` ([mapping of the main binary](https://github.com/google/pprof/blob/91b7bce49751dd18be082ed4aa210d63ae53a771/proto/profile.proto#L58-L60)) if it is available and valid (e.g is not "[vdso]");
      * Use `service_name` label if it is available (I'm not too sure if it should take precedence over `Mapping.filename` or not);
      * Use "unknown" otherwise. 
    * The field is an index into the strings table.
    * The row order of the `profiles.parquet` table does not change (ordered by series ID);
    * Ensure backward compatibility.
 - [x] (@kolesnikovae) Implement the wire format for the new symbols db;
 - [x] (@kolesnikovae) Implement in-memory representation of the db;
 - [x] (@cyriltovena) Integrate into phlaredb:
    - [x] Write path
    - [x] Read path
       * [x] In-memory
       * [x] File
       * [x] Ensure backward compatibility
    - [x] Block flush

PR #802, opened 2023-06-27
Fix backwards compatibility with block version 1

simonswine:
#767 broke compatibility with version 1 blocks, this fixes the two issues and also adds sample blocks and tests to ensure compatibility.


cyriltovena:
Did you figure why it broke ?

pkg/phlaredb/block_querier.go
```diff
@@ -983,17 +986,29 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
-	pIt := query.NewJoinIterator(
-		0,
-		[]query.Iterator{
+
+	var (
+		buf       [][]parquet.Value
+		joinIters []query.Iterator
+	)
+
+	if b.meta.Version >= 2 {
```
cyriltovena, line 995: I think I want to include total sample in v2 WDYT ?

cyriltovena (approved):
LGTM

pkg/phlaredb/block_querier.go
simonswine: I would really throw around with versions and make that version+1, also the sorted should be a version+1

pkg/phlaredb/query/repeated.go
```diff
@@ -83,8 +84,10 @@ func (it *repeatedPageIterator[T]) seekRowNum() int64 {
 		return i.RowNumber()
 	case int64:
 		return i
+	case uint32:
+		return int64(i)
```
simonswine, line 88: Problem #2, panic with uint32s

pkg/phlaredb/block_querier.go
```diff
@@ -983,17 +986,29 @@ func (b *singleBlockQuerier) SelectMatchingProfiles(ctx context.Context, params
...
+		joinIters = []query.Iterator{
 			b.profiles.columnIter(ctx, "SeriesIndex", newMapPredicate(lblsPerRef), "SeriesIndex"),
 			b.profiles.columnIter(ctx, "TimeNanos", query.NewIntBetweenPredicate(model.Time(params.Start).UnixNano(), model.Time(params.End).UnixNano()), "TimeNanos"),
 			b.profiles.columnIter(ctx, "StacktracePartition", nil, "StacktracePartition"),
```
simonswine, line 999: Problem #1: This requires the column to be there, maybe should have a orNull predicate or something.

Issue #800, opened 2023-06-26
Sort Rows by Series when flushing to disk


cyriltovena:
Currently we don't sort across row groups of profiles when flushing all of them.

 I believe we should be able to stream all row groups from https://github.com/grafana/phlare/blob/b67c811f486955041422a923177545cbce024471/pkg/phlaredb/profile_store.go#L360 and reorder them by SeriesID then timestamp.

simonswine:
This will improve data locality a ton, but I am a bit unsure how this will impact querying, as the order of querying will be:

* Timestamp first then SeriesID

And blocks will be strictly stored in
* SeriesID first then Timestamp

Currently the sorting is more like:
- Within a single row group strictly: Series first then Timestamp
- Across row groups, loosely timestamp ordered

If we query a time range only impacting a part of the ~3 hours within a block, we could get away by only reading pages that fall within the time ranges (based on the pages Min/Max). With this change we can only access pages by their SeriesIDs min/max.  I don't expect a major issue, I am just wondering.

Fairly relevant to the change in #799

simonswine:
Maybe we need to address our query behaviour to use the order that is in the blocks as well.

PR #803, opened 2023-06-27
Sort Profiles across row groups when flushing blocks

cyriltovena:
Fixes https://github.com/grafana/phlare/issues/800

This k-way merge rowgroups on disk when we flush to a full block. Which means now a single series is expected to appear in only on rowgroup.

The only caveat if we only cut the final file rowgroup per rowcount but I think we can live with this since we never cut by size.

### DIFF
--- a/pkg/parquet/row_reader.go
+++ b/pkg/parquet/row_reader.go
@@ -1,6 +1,7 @@
 package parquet
 
 import (
+	"fmt"
 	"io"
 
 	"github.com/grafana/phlare/pkg/iter"
@@ -45,8 +46,8 @@ func NewMergeRowReader(readers []parquet.RowReader, maxValue parquet.Row, less f
 		return readers[0]
 	}
 	its := make([]iter.Iterator[parquet.Row], len(readers))
-	for i, r := range readers {
-		its[i] = NewBufferedRowReaderIterator(r, defaultRowBufferSize)
+	for i := range readers {
+		its[i] = NewBufferedRowReaderIterator(readers[i], defaultRowBufferSize)
 	}
 
 	return &MergeRowReader{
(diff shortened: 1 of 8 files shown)


pkg/phlaredb/profile_store_test.go
```diff
@@ -169,41 +169,41 @@ func TestProfileStore_RowGroupSplitting(t *testing.T) {
...
-			values:          sameProfileStream,
-		},
-		{
-			name:            "a stream ending after half of the samples and a new one starting",
-			cfg:             &ParquetConfig{MaxRowGroupBytes: 1828, MaxBufferRowCount: 100000},
-			expectedNumRGs:  10,
-			expectedNumRows: 100,
-			values:          profileStreamEndingAndStarting(50),
-		},
+		// {
```
cyriltovena, line 172: Still need to figure those.

cyriltovena:
This is already running in dev everything looks good 
image

 Seems like finding values in using row.Range is the new CPU additions. Nothing problematic but I want to see if we can do better.

cyriltovena:
I was able to speed this up quite significantly 

```
name            old time/op    new time/op    delta
ProfileRows-16     310ns ± 3%      31ns ± 5%  -90.05%  (p=0.008 n=5+5)

name            old alloc/op   new alloc/op   delta
ProfileRows-16     0.00B          0.00B          ~     (all equal)

name            old allocs/op  new allocs/op  delta
ProfileRows-16      0.00           0.00          ~     (all equal)
```

pkg/parquet/row_reader.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	r.buff = r.buff[:r.bufferSize]
+	n, err := r.reader.ReadRows(r.buff)
+	if n == 0 {
+		return false
+	}
+	if err != nil && err != io.EOF {
+		r.err = err
+		return false
+	}
```
kolesnikovae, line 116: Maybe we should check error before `n == 0` to not miss it

kolesnikovae (approved):
LGTM!

pkg/parquet/row_reader.go
cyriltovena: yeah good point

PR #804, opened 2023-06-28
BlockQuerier.SelectMatchingProfiles: Fix condition that caused memory wastage

simonswine:
While investigating grafana/pyroscope#2038, I noticed we never had more than one slice element and as we allocate 100 elements, that added 100  times more memory pressure than necessary.

Screenshot from 2023-06-28 10-21-51


cyriltovena (approved):
LGTM

PR #807, opened 2023-06-30
Improve block querying with a binary join and sync iterator

simonswine:
This should avoid over reading at then end of the join (with Tempo's sync iterator) and also make sure it seeks before looking into the second iterator of the join. (Binary Join iterator).


pkg/phlaredb/query/iters.go
```diff
@@ -96,17 +95,188 @@ func (t RowNumber) Valid() bool {
...
-	// New children up through the definition level
-	for i := repetitionLevel + 1; i <= definitionLevel; i++ {
-		t[i] = 0
-	}
-
-	// Children past the definition level are undefined
-	for i := definitionLevel + 1; i < len(t); i++ {
-		t[i] = -1
+	// the following is nextSlow() unrolled
+	switch repetitionLevel {
```
cyriltovena, line 101: Do you know what's going on here ? why 5 ?

simonswine: I think might be the highest definition level the otel schema has? 

Probably @joe-elliott might be able to anwser that better.

In the grand scheme of things, the gain is probably not worth the unrolling for us, happy to put that back to the original (https://github.com/grafana/tempo/pull/2412).

joe-elliott: @simonswine is correct. it's just an arbitrary depth that is enough for our schema.

i don't recall the numbers but profiling heavy "needle in the haystack" queries showed a large %age of the time in this function (which is why we unrolled it).

cyriltovena: I'm fine if we don't overflow our current definition/repetition level for the 3 column we use this iterator with.

pkg/phlaredb/query/iters_test.go
```diff
@@ -394,20 +409,35 @@ func TestBinaryJoinIterator(t *testing.T) {
...
 			expectedResultCount: 2 * 8, // expect two profiles for each series
 			timePredicate:       NewIntBetweenPredicate(0, 1000),
+			seriesPageReads:     1,
+			timePageReads:       1,
 		},
 		{
 			name:                "time before results",
 			expectedResultCount: 0,
 			timePredicate:       NewIntBetweenPredicate(-10, -1),
+			seriesPageReads:     1,
```
simonswine, line 419: @cyriltovena We could prevent this  single read if we would push down TimeNanos predicates into the SeriesID iterator. 

I don't think we should do this as that PR though 😆

pkg/phlaredb/query/iters.go
simonswine: It is inline with limit  imposed by `RowNumber [6]int64§

simonswine:
The difference in the `ColumnIterator` vs `SyncIterator` is minuscule:

```
name                    old time/op    new time/op    delta
ColumnIterator/sync-12    17.9ms ± 5%    17.4ms ± 4%  -3.14%  (p=0.007 n=10+10)

name                    old alloc/op   new alloc/op   delta
ColumnIterator/sync-12    68.6MB ± 0%    65.7MB ± 0%  -4.25%  (p=0.000 n=10+10)

name                    old allocs/op  new allocs/op  delta
ColumnIterator/sync-12      200k ± 0%      200k ± 0%  -0.09%  (p=0.000 n=10+9)
```

But getting rid of the async aspect is beneficial stopping overreading.

A bigger impact is on the `JoinIterator` vs `BinaryJoinIterator` and the amount of pages read from a block:
- It will now seekOrNext as soon as possible rather than reading the first page of every column and the move on.
- Also it is quite beneficial doing a next instead of sync if the it is the adjacent row (that might be something we would rather have in the SyncIterator maybe)

Left --- / Right +++

image


The impact to the overall query time is no longer so significant, this was probably due a bug that missed out on some matches.

pkg/phlaredb/query/iters.go
```diff
@@ -223,459 +386,112 @@ var columnIteratorResultPool = sync.Pool{
...
-		if j.peeks[iterNum] == nil || CompareRowNumbers(to.DefinitionLevel, j.peeks[iterNum].RowNumber, to.RowNumber) == -1 {
-			columnIteratorResultPoolPut(j.peeks[iterNum])
-			if iter.Seek(to) {
-				j.peeks[iterNum] = iter.At()
-			} else {
-				j.peeks[iterNum] = nil
-			}
+		} else {
+			// the right value can't be smaller than the left one because we seeked beyond it
+			panic("not expected to happen")
```
cyriltovena, line 471: Can you make the panic message more specific ?

cyriltovena (approved):
LGTM

PR #810, opened 2023-07-04
Update go-groupvarint dependency

kolesnikovae:
See https://github.com/dgryski/go-groupvarint/pull/4 for details

Thanks @simonswine!


cyriltovena (approved):
LGTM

PR #811, opened 2023-07-04
Use DoUntilQuorumWithoutSuccessfulContextCancellation instead of DoUntilQuorum

cyriltovena:
DoUntilQuorum has changed in behaviour and by default cancels all context even succesful one.

We should use `DoUntilQuorumWithoutSuccessfulContextCancellation` instead which makes it obvious we don't want cancellation for successful ones.


simonswine (approved):
LGTM

PR #812, opened 2023-07-04
Fixes arm32 build while using dskit

cyriltovena:
vendor in https://github.com/grafana/dskit/pull/333 to fix arm32 build


simonswine (approved):
LGTM

Issue #797, opened 2023-06-26
Improve Locations in memory representation


cyriltovena:
Same as https://github.com/grafana/phlare/pull/796 we can have reduce memory usage drastically if we have a better representation in memory that is irrespective of the file format.

The same  inefficient format is also used when opening a block and cache in memory.

<img width="2461" alt="image" src="https://github.com/grafana/phlare/assets/1053421/ba13d542-82ba-4bc0-9021-bf847a30ad29">

kolesnikovae:
Resolved via #814

PR #814, opened 2023-07-06
Improve memory layout of locations, functions, and mappings

kolesnikovae:
Addresses https://github.com/grafana/phlare/issues/797

The numbers below is an estimate of the memory consumption reduce:
```
pyroscope_head_size_bytes{type="functions"} 240 -> 72
pyroscope_head_size_bytes{type="locations"} 344 -> 152
pyroscope_head_size_bytes{type="mappings"}  192 -> 96
```

It is expected that the overall memory consumption of the cluster will be reduced by appx. 20%.


pkg/phlaredb/schemas/v1/locations.go
```diff
@@ -27,15 +23,53 @@ func (*LocationPersister) SortingColumns() parquet.SortingOption {
...
+	// The id of the corresponding profile.Mapping for this location.
+	// It can be unset if the mapping is unknown or not applicable for
+	// this profile type.
+	MappingId uint32
+	// Provides an indication that multiple symbols map to this location's
+	// address, for example due to identical code folding by the linker. In that
+	// case the line information above represents one of the multiple
+	// symbols. This field must be recomputed when the symbolization state of the
+	// profile changes.
+	IsFolded bool
```
cyriltovena, line 58: struct padding should be improved

kolesnikovae: There are two changes comparing to the proto definition:
1. `MappingId` type uint64 -> uint32;
2. `IsFolded` moved next to `MappingId`, to avoid extra-padding.

The current type size is 48 bytes, given the 8-bytes alignment. Further improvement does not seem to be possible.

cyriltovena (approved):
LGTM

simonswine (approved):
LGTM. Nice work, this will help our memory headroom 🙂

PR #816, opened 2023-07-06
Fix makefile issues

simonswine:
- Correct environment variables GOOS GOOARCH: Was not setup correctly
- Give empty second argument: Avoids a warning


cyriltovena (approved):
LGTM

PR #817, opened 2023-07-06
Reduce max duration for kind deploys

simonswine:
This allows better testing for blocks on ingester local and bucket storage


cyriltovena (approved):
LGTM

PR #828, opened 2023-07-07
Ensure we check iterators errors when deduplicating

cyriltovena:
I suspect this could be the cause for couple of missing errors.

### DIFF
--- a/pkg/querier/select_merge.go
+++ b/pkg/querier/select_merge.go
@@ -261,7 +261,9 @@ func skipDuplicates(ctx context.Context, its []MergeIterator) error {
 	}
 	span.LogFields(otlog.Int("duplicates", duplicates))
 	span.LogFields(otlog.Int("total", total))
-
+	for _, it := range its {
+		errors.Add(it.Err())
+	}
 	return errors.Err()
 }
 
--- a/pkg/querier/select_merge_test.go
+++ b/pkg/querier/select_merge_test.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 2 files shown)


kolesnikovae (approved):

pkg/querier/select_merge.go
```diff
@@ -261,7 +261,9 @@ func skipDuplicates(ctx context.Context, its []MergeIterator) error {
 	}
 	span.LogFields(otlog.Int("duplicates", duplicates))
 	span.LogFields(otlog.Int("total", total))
-
+	for _, it := range its {
+		errors.Add(it.Err())
```
simonswine, line 265: Not fully sure if that will really cover all bases, in theory Err needs to be called every time `Next()` of the particular iterator is `false`. So that probably should happen as part of `moveNext()` and kept in the tree struct.

Still better than having no errors at all. 🙂

simonswine (approved):
LGTM

pkg/querier/select_merge.go
cyriltovena: Yeah this is a very good point. I think We can fix that too let me see.

cyriltovena: Checking for each iterators errors on a false next might be expensive if we do this outside of the loser tree.

I'm wondering if we should change the tree to be aware of Err() which it is not right now. @bboreham wanted to ultimately have the same library for all project, but some doesn't have this Err() and this could impact performance.

I'm inclined to have two version and may be use that new version with Err() management in Pyroscope.

WDYT ?

simonswine: Yes I think it is worth having the extra check. I guess at least Next is most times `true`, before it is once `false` for every iterator.

cyriltovena: I'll replace that PR with another one then

PR #829, opened 2023-07-07
Add app selector search

Rperry2174:
Adds search to app selector bar... 

Warning -- this was a [chatgpt generated change](https://chat.openai.com/share/aff13004-9e18-4630-aeac-13c3f3290a35) but seems to be looking / working fine or at least better than having no search at all. 

https://github.com/grafana/phlare/assets/23323466/44338f49-904f-46ad-89b1-0d5e8cd77e95


cyriltovena (approved):
LGTM

Love it, we should add it to the plugin too.

PR #831, opened 2023-07-08
Terminate stream in store gateway if no blocks found

kolesnikovae:
Currently, the query fails with a cryptic error when object store is queried before the first block is uploaded. The problem is that the querier expects the stream to the store-gateway is properly utilised and terminated regardless of anything, however store-gateways leave the stream intact, if no suitable blocks found. In ingesters, the stream is always utilised, even if there is no suitable data (subject of another issue, see grafana/pyroscope#2047). I decided to not change the logic but instead terminate the stream in the same way it is done when actual data is streamed.


cyriltovena (approved):
LGTM

PR #836, opened 2023-07-11
Adds iteration error support in the loser.Tree

cyriltovena:
This makes loser.Tree aware of Err() and stops as soon as an iterator return false from Next() with an Err() != nil.

Supersedes https://github.com/grafana/phlare/pull/828


kolesnikovae:
LGTM

kolesnikovae (approved):

Issue #688, opened 2023-05-12
Loss of data on restart


kolesnikovae:
A restart may result in losing the head block:
![image](https://github.com/grafana/phlare/assets/12090599/41413187-f2bc-4c62-b4bb-c0112700937d)

Log from one of the ingesters (cluster of 4 ingesters with the default replication factor): apparently, shutdown didn’t finish gracefully:
![image](https://github.com/grafana/phlare/assets/12090599/2838bd21-ccbf-44e5-ab5f-8c846efe294c)

I find it interesting that none of the ingesters has finished shutdown as well: no panics or explicit errors in the log – however, most likely this indicates a problem with log collection.

It seems that under some circumstances a block is not picked up after the restart, and the head is getting polluted with chunks:
```
/ $ ls -la /data/1218/head | wc -l
276
/ $ du -sh /data/1218/head
49.9G   /data/1218/head
```

kolesnikovae:
Also see https://github.com/grafana/pyroscope/issues/2076

cyriltovena:
Possible we need to increase the grace period

```
    deployment.mixin.spec.template.spec.withTerminationGracePeriodSeconds(4800)

```

kolesnikovae:
Termination grace period added in grafana/phlare#837. Closing for now

PR #837, opened 2023-07-11
Add terminationGracePeriodSeconds helm value

kolesnikovae:
It's been reported multiple times that the unfinished block is lost at shutdown. One of the most common reasons in case of k8s deployment is the fact that ingesters may not be able to flush all the data before being terminated forcibly due to expiration of `terminationGracePeriodSeconds` (defaults to 30s).

The change makes the parameter configurable: now it is set to 10m by default for ingesters when deployed as micro services.

Related #688

### DIFF
--- a/operations/phlare/helm/phlare/Chart.yaml
+++ b/operations/phlare/helm/phlare/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: phlare
 description: 🔥 horizontally-scalable, highly-available, multi-tenant continuous profiling aggregation system
 type: application
-version: 0.5.3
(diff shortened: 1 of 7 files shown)


cyriltovena (approved):
LGTM

simonswine (approved):
LGTM! Yes that is a big oversight

simonswine:
If you increment the version number it should auto release the chart:

https://github.com/grafana/phlare/blob/0ba6fc4043259e69c7305cd66b9c02505640e3b8/operations/phlare/helm/phlare/Chart.yaml#L5

PR #839, opened 2023-07-11
Measure page read duration

kolesnikovae:
Despite the fact that we can derive the information based on the "distance" between events, is might be inaccurate:

image


cyriltovena (approved):
LGTM

PR #842, opened 2023-07-11
Correctly saves the buffer capacity in BufferedRowReaderIterator

cyriltovena:
I was looking at compaction and trying to figure why it was slow.

```
❯ benchstat before.txt afte.txt
name                  old time/op    new time/op    delta
BufferedRowReader-16     709ns ±14%     306ns ± 3%   -56.88%  (p=0.008 n=5+5)

name                  old alloc/op   new alloc/op   delta
BufferedRowReader-16    2.72kB ± 0%    0.00kB       -100.00%  (p=0.008 n=5+5)

name                  old allocs/op  new allocs/op  delta
BufferedRowReader-16      1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
```

This is most likely makes flushing a block 50% faster and lighter.

PS: Not fully sure why we loose the slice capacity when slicing.

### DIFF
--- a/pkg/parquet/row_reader.go
+++ b/pkg/parquet/row_reader.go
@@ -84,7 +84,10 @@ func (s *MergeRowReader) ReadRows(rows []parquet.Row) (int, error) {
 }
 
 type BufferedRowReaderIterator struct {
-	reader     parquet.RowReader
+	reader       parquet.RowReader
+	bufferedRows []parquet.Row
+
+	// buff keep the original slice capacity to avoid allocations
 	buff       []parquet.Row
 	bufferSize int
 	err        error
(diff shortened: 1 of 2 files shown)


pkg/parquet/row_reader.go
```diff
@@ -100,16 +103,16 @@ func NewBufferedRowReaderIterator(reader parquet.RowReader, bufferSize int) *Buf
 }
 
 func (r *BufferedRowReaderIterator) Next() bool {
-	if len(r.buff) > 1 {
-		r.buff = r.buff[1:]
+	if len(r.bufferedRows) > 1 {
+		r.bufferedRows = r.bufferedRows[1:]
```
simonswine, line 107: That old line surely reduced the capacity by`-1`


It is only when you cut of the end that the capacity stays the same, so `buf[:cap(buff)-1]`, will still have the same capactity.

simonswine (approved):
LGTM

cyriltovena:
The PR is ready to merge but it trigger another bug hidden in the merge. I'll update this PR to have the fix soon.

PR #844, opened 2023-07-11
Allow to filter labelNames in series call

simonswine:
This will improve the response size of the application selector in the pyroscope app:

Can be used like this:

```
curl -vL  -d '{"label_names":["__profile_type__","__service_name__"]}' -X POST -H "content-type: application/json" http://localhost:4100/querier.v1.QuerierService/Series

{
  "labelsSet": [
    {
      "labels": [
        {
          "name": "__profile_type__",
          "value": "block:contentions:count:contentions:count"
        },
        {
          "name": "__service_name__",
          "value": "pyroscope"
        }
      ]
    },
[...]
```


cyriltovena:
Nice can you test this in ops ? See how long it takes, I think gzip is a must here since it will have a lot of profile_types repeated

simonswine:
I have deployed this to ops, not I am on some not too fast wifi:

```
# not compressed and only __profile_types__ + __service_name__, trace id 949237ad6c92c78
100  366k    0  366k  100    55  98547     14  0:00:03  0:00:03 --:--:-- 98652
# gzip and only _profile_types__ + __service_name__, trace id 6dbd52a0fb25c754
100 14378    0 14323  100    55   4417     16  0:00:03  0:00:03 --:--:--  4440

# Gzip all labels, trace ID 1163c38219238509
100 3396k    0 3396k    0     2   303k      0 --:--:--  0:00:11 --:--:--  860k
# No gzip all labels, timed out after 30 sec+, trace ID 50a1f91b7db66dfa
100 58.9M    0 58.9M    0     2  1732k      0 --:--:--  0:00:34 --:--:-- 2899k
```

To get gzip with connect we need to just supply the `accept-encoding: gzip`

cyriltovena (approved):
LGTM

PR #845, opened 2023-07-12
Fixes iterator At value re-used

cyriltovena:
This is a refactoring a bit how we deal with RowReader but also fixes a bug where we were re-used the `At` of the inner iterator inside the RowReader.

Currently causing no grief but did when working on compaction. Already include https://github.com/grafana/phlare/pull/842.

Closes https://github.com/grafana/phlare/pull/842

### DIFF
--- a/pkg/parquet/row_reader.go
+++ b/pkg/parquet/row_reader.go
@@ -168,7 +168,9 @@ func ReadAllWithBufferSize(r parquet.RowReader, bufferSize int) ([]parquet.Row, 
 			return rows, err
 		}
 		if n != 0 {
-			rows = append(rows, batch[:n]...)
+			for i := range batch[:n] {
+				rows = append(rows, batch[i].Clone())
+			}
 		}
 		if n == 0 || err == io.EOF {
 			break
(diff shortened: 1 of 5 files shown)


pkg/parquet/row_reader.go
```diff
@@ -70,21 +81,27 @@ func (s *MergeRowReader) ReadRows(rows []parquet.Row) (int, error) {
...
+			runutil.CloseWithLogOnErr(util.Logger, it.Iterator, "failed to close iterator")
+			if it.Err() != nil {
+				return n, it.Err()
 			}
 			return n, io.EOF
 		}
-		rows[n] = s.tree.Winner().At()
+		rows[n] = rows[n][:0]
+		for _, c := range it.At() {
+			rows[n] = append(rows[n], c.Clone())
```
cyriltovena, line 93: We have to clone each value here since At can change the current row.

kolesnikovae (approved):
LGTM

cyriltovena:
Well thanks to this PR found another bug :)

PR #847, opened 2023-07-12
fix(ebpf): procmap parse test fix

korniltsev:
Use same packed dev structure in tests as in implementation

We use mkdev result as a cache key, it does not matter what the implementation is, it just should be the same


cyriltovena (approved):
LGTM

PR #848, opened 2023-07-12
Optimise in-memory parquet reader

kolesnikovae:
The change makes in-memory parquet reader to not use the stored schema for reconstruction via `reflect`, and use hand-written reconstruction instead.

### DIFF
--- a/pkg/phlaredb/block_querier.go
+++ b/pkg/phlaredb/block_querier.go
@@ -294,7 +294,7 @@ type singleBlockQuerier struct {
 	openLock    sync.Mutex
 	opened      bool
 	index       *index.Reader
-	strings     inMemoryparquetReader[*schemav1.StoredString, *schemav1.StringPersister]
+	strings     inMemoryparquetReader[string, *schemav1.StringPersister]
 	functions   inMemoryparquetReader[*schemav1.InMemoryFunction, *schemav1.FunctionPersister]
 	locations   inMemoryparquetReader[*schemav1.InMemoryLocation, *schemav1.LocationPersister]
 	mappings    inMemoryparquetReader[*schemav1.InMemoryMapping, *schemav1.MappingPersister]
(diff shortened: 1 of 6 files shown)


pkg/phlaredb/block_querier.go
```diff
@@ -1216,15 +1216,26 @@ func (r *inMemoryparquetReader[M, P]) open(ctx context.Context, bucketReader phl
...
 
 	// read all rows into memory
-	r.cache = make([]M, r.file.NumRows())
+	r.cache = make([]M, 0, r.file.NumRows())
 	var read int
 	for _, rg := range r.file.RowGroups() {
-		err := func(rg parquet.RowGroup) error {
-			reader := parquet.NewGenericRowGroupReader[M](rg)
-			defer runutil.CloseWithLogOnErr(util.Logger, reader, "closing parquet generic row group reader")
+		err = func(rg parquet.RowGroup) error {
```
cyriltovena, line 1222: I suspect we should also run this in parallel ?

kolesnikovae: I'm not sure. The problem is that it would cost us lots of memory, because [page reader acqiures a buffer](https://github.com/segmentio/parquet-go/blob/main/file.go#L519) (of `ReadBufferSize` size 2MB) per column chunk, and being read concurrently, the internal buffer pool would be useless.

Interestingly, I see no apparent reason why column chunks should not be read concurrently: https://github.com/segmentio/parquet-go/blob/5d42db8f0d4728c31759068f08da15df44c6cc7f/row_group.go#L320, since buffers are already allocated, it should not be this wasteful

pkg/phlaredb/schemas/v1/strings.go
```diff
@@ -11,40 +11,24 @@ var stringsSchema = parquet.NewSchema("String", phlareparquet.Group{
 	phlareparquet.NewGroupField("String", parquet.Encoded(parquet.String(), &parquet.RLEDictionary)),
 })
 
-type StoredString struct {
-	ID     uint64 `parquet:",delta"`
-	String string `parquet:",dict"`
-}
```
kolesnikovae, line 17: I didn't manage to find how the ID field is used

cyriltovena (approved):
LGTM

pkg/phlaredb/block_querier.go
cyriltovena: yeah column are read in parallel in that case, but I thought it would be good to process multiple rowgroup at once, not sure if it would help as you say let's see.

PR #849, opened 2023-07-12
Refactor block meta iteration

cyriltovena:
Move block iteration to `block` package so that we can reuse it.

Doesn't work on filesystem for now. see https://github.com/grafana/pyroscope/issues/2029


simonswine (approved):
LGTM

PR #852, opened 2023-07-12
Fix BinaryJoinIterator.Seek

simonswine:
- Implement BinaryJoinIterator.Seek() properly
- Fix nextOrSeek entry conditions


cyriltovena (approved):
LGTM

PR #855, opened 2023-07-13
Fix linting error in frontend

simonswine:
I have no idea what I am doing, but this should fix the linting issue we see on every PR.

### DIFF
--- a/public/app/overrides/components/Toolbar.tsx
+++ b/public/app/overrides/components/Toolbar.tsx
@@ -14,7 +14,7 @@ import Button from '@webapp/ui/Button';
 import LoadingSpinner from '@webapp/ui/LoadingSpinner';
 import DateRangePicker from '@webapp/components/DateRangePicker';
 import RefreshButton from '@webapp/components/RefreshButton';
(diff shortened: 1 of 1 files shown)


public/app/overrides/components/Toolbar.tsx
```diff
@@ -14,7 +14,7 @@ import Button from '@webapp/ui/Button';
 import LoadingSpinner from '@webapp/ui/LoadingSpinner';
 import DateRangePicker from '@webapp/components/DateRangePicker';
 import RefreshButton from '@webapp/components/RefreshButton';
-import AppSelector from '@phlare/components/AppSelector/AppSelector';
+import { AppSelector } from '@phlare/components/AppSelector/AppSelector';
```
darrenjaneczek, line 17: It's funny because both are technically valid.

```
import { AppSelector } from '@phlare/components/AppSelector/AppSelector';
```

specifies only importing `AppSelector`, but since `AppSelector` is declared as the default export in that module, what was originally there is functionally equivalent.

PR #857, opened 2023-07-13
Use labels API when possible instead of Series

cyriltovena:
Series API with `{}` query is expensive.


cyriltovena:
I want to finish this and get rid of any old api too

cyriltovena:
Alright so now we only use connect-go API. (Series, LabelNames and LabelsValues)

On top of this we also pass the current query as matchers to correctly reduce values and names for the selected workload.

public/app/overrides/services/appNames.ts
```diff
@@ -1,38 +0,0 @@
...
-      return {
-        name,
-        spyName: 'gospy',
-        units: 'unknown',
-      };
-    });
-  })
-);
-
-export async function fetchApps(): Promise<
```
cyriltovena, line 26: Couldn't find where this is called.

cyriltovena:
See below it shows only 4 pods of the current app.
image

public/app/overrides/services/appNames.ts
cyriltovena: It's not even loaded.

simonswine (approved):
LGTM

I have noticed that the app menur no longer closes when you select an app and profile type.

Also when selecting an app with more matches (e.g. cortex-dev-01/ingester), I am getting this when I select tags fairly often (needs like 10-20 tries)

Screenshot 2023-07-17 at 13 05 23

simonswine:
I see that that was on auto merge 🙂 

Unsure what you mean with the 4 pods?

I am able to show more:

Screenshot 2023-07-17 at 13 06 17

cyriltovena:
> I see that that was on auto merge 🙂
> 
> Unsure what you mean with the 4 pods?
> 
> I am able to show more:
> 
> Screenshot 2023-07-17 at 13 06 17

I meant that before it would show all pods irrespectively of the selection application.

PR #858, opened 2023-07-13
Fix app selector scrolling

Rperry2174:
This takes the:
- [app selector from grafana/pyroscope](https://github.com/grafana/pyroscope/blob/ce966012e5ada7d4d27d7f07a6261ff11433d955/webapp/javascript/components/AppSelector/index.tsx)
- [app selector from grafana/phlare](https://github.com/grafana/phlare/blob/64da02ed79fdaa09b75ed34f45e63b18665fc42d/public/app/components/AppSelector/AppSelector.tsx) (which overrides the grafana/pyroscope version to accomodate the different datastructure that exists in phlare, but not pyroscope

And then I asked for some help from [chatgpt](https://chat.openai.com/share/289269ce-43a8-48f8-a597-8a2117c531d9) to bring some functionality from the pyroscope version into the phlare version.

Had to make some tweaks after, but now it looks better. Would like to merge this and get it into the next weekly release for phlare so we can then add it to the Grafana cloud profiles [app plugin](https://github.com/grafana/pyroscope-app-plugin/blob/main/package.json#L83) which imports the ui from phlare. 

Fixes app selector scrolling

https://github.com/grafana/phlare/assets/23323466/4108119b-7f09-47d4-81f1-7b5ae83f022c

### DIFF
--- a/public/app/components/AppSelector/AppSelector.tsx
+++ b/public/app/components/AppSelector/AppSelector.tsx
@@ -149,21 +149,12 @@ export const SelectorModalWithToggler = ({
 
   const groups = useMemo(() => {
     const allGroups = leftSideApps.map((app) => app.name.split('-')[0]);
-
-    // console.log('leftSideApps', leftSideApps);
-    // console.log('allGroups', allGroups);
-
-
     const uniqGroups = Array.from(new Set(allGroups));
 
-    // console.log('uniqGroups', uniqGroups);
-  
     const dedupedUniqGroups = uniqGroups.filter((x) => {
       return !uniqGroups.find((y) => x !== y && y.startsWith(x));
     });
 
-    // console.log('dedupedUniqGroups', dedupedUniqGroups);
-  
     const groupOrApp = dedupedUniqGroups.map((groupName) => {
       const appNamesEntries = leftSideApps.filter((app) =>
         app.name.startsWith(groupName)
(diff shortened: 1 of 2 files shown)


public/app/components/AppSelector/AppSelector.tsx
```diff
@@ -134,6 +147,54 @@ export const SelectorModalWithToggler = ({
     return selectedApp?.__profile_type__ === a.__profile_type__;
   };
 
+  const groups = useMemo(() => {
+    const allGroups = leftSideApps.map((app) => app.name.split('-')[0]);
```
Rperry2174, line 151: I guess here we should maybe use `service_name` and simply calculate the height of the dropdown based off the number of `service_name`.

If I'm understanding correctly the idea here is to:
1. Calculate the necessary height of the modal based off the number of "groups" that will exist
2. If the modal is bigger than the screen then set the modal to slightly bigger height than the screen... if the modal is smaller than the screen then just adapt to the size of the modal

right now the logic is assuming that the "groups" are differentiated by the text before the first `-` but thats not true. I guess they are now based off of the `service_name`? 

cc @korniltsev  @cyriltovena  @petethepig

Rperry2174: this isn't super helpful because I just cheated so I could make the dropdown very high by adding random numbers to each `name`, but for production data, for example, will the height of the modal will be determined by the unique number of `service_names`

petethepig (approved):
LGTM, `max-height` would be a slightly cleaner solution but this will work too.

PR #859, opened 2023-07-14
fix(ebpf): vendor rlimit package from cilium, to avoid bpf syscalls during init

korniltsev:
- added ebpf/cmd/playground which is a convenient way to run/test/develop profiler outside grafana agent. this is not relevant to the PR, I can remove it and submit as a separate PR
- update cilium/ebpf to v11 it fixes one bpf syscall during cilium/ebpf/btf init 
- copy the cilium/ebpf/rlimit package with the following changes

    1.  haveMemcgAccounting is initialized once during RemoveMemlock
    2. bunch of internal dependencies copied to rlimit/internal/internal_mirror.go

Alternative to this PR would be a fork of cillium, which I'd like to avoid if possible (here's the fork https://github.com/grafana/ebpf/pull/1)

Upstream issue https://github.com/cilium/ebpf/issues/1096


petethepig (approved):
LGTM

PR #866, opened 2023-07-18
Add flamegraph export feature (PNG/JSON/PPROF)

cyriltovena:
I will add flamegraph.com later, because I want to use CORS so that anyone can use flamgraph.com without a proxy.

Pprof is only available for single view, but returns the full merge of pprof, including mapping, lines and filenames.

### DIFF
--- a/public/app/overrides/components/ExportData.tsx
+++ b/public/app/overrides/components/ExportData.tsx
@@ -1,15 +1,434 @@
-import React from 'react';
+/* eslint-disable react/destructuring-assignment */
+import React, { useState } from 'react';
+import { format } from 'date-fns';
+import OutsideClickHandler from 'react-outside-click-handler';
+import { Tooltip } from '@pyroscope/webapp/javascript/ui/Tooltip';
+import Button from '@webapp/ui/Button';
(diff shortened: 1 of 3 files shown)


cyriltovena:
image

public/app/overrides/components/ExportData.tsx
```diff
@@ -1,15 +1,434 @@
...
+  };
+
+  const downloadPprof = async function () {
+    if (!exportPprof) {
+      return;
+    }
+
+    if (props.exportPprof) {
+      // get user input from modal
+      const customExportName = await getCustomExportName('profile.pb.gz');
```
simonswine, line 274: I wonder if we could default to something better like from date

```
${YYYY-MM-DD HH:MM:SS}-${name}-${matchers}
```

My download folder already has profile.pb.gz 🙂

simonswine (approved):
LGTM 

That all seemed to work for me! Let's merge before the deadline

public/app/overrides/components/ExportData.tsx
cyriltovena: Yes we can !

